PR #72, opened 2021-12-02
Fix: Revert opcode

ArthurBraud:
Fix Revert

### DIFF
--- a/docs/opcodes/FD.mdx
+++ b/docs/opcodes/FD.mdx
@@ -5,18 +5,30 @@ group: System operations
 
 ## Notes
 
-Equivalent to [CREATE](/#F0), except the salt allows to create at a deterministic address.
+Abort execution, undo all state changes, return a value and refund the remaining gas to the caller.
+The error message returned is available to the caller in the returndata buffer.
 
 ## Stack input
 
-0. `Integer`: value in [wei](https://www.investopedia.com/terms/w/wei.asp) to send to the new account.
-1. `Integer`: offset in the [memory](/about) in bytes, the instructions of the new account.
-2. `Integer`: how many bytes to copy (size of the instructions).
-3. `Integer`: salt, 32-byte value used to create the new account at a deterministic address.
+0. `Integer`: offset in the [memory](/about) in bytes. The starting position of the value to return.
+1. `Integer`: length in the [memory](/about) in bytes to return.
+
+## Returndata output
+
+memory[offset:offset+length]
+
+## Examples
+
+| Memory |
+|--------|
+| `0xFF00000000000000000000000000000000000000000000000000000000000000` |
+
+|  Input   | Output |
+|---------:|-------:|
+| `0x0001` | `0xFF` |
 
 ## Error cases
 
 The transaction is reverted in those cases:
 - Not enough gas.
 - Not enough values on the stack.
-- The current execution context is from a [STATICCALL](/#FA).
--- a/docs/opcodes/FD.mdx
+++ b/docs/opcodes/FD.mdx
@@ -5,27 +5,27 @@ group: System operations
 
 ## Notes
 
-Abort execution, undo all state changes, return a value and refund the remaining gas to the caller.
-The error message returned is available to the caller in the returndata buffer.
+Stop the current context execution, revert the state changes (see [STATICCALL](/#FA) for a list of state changing opcodes) and return unused gas to the caller. If the execution is stopped with REVERT, the value 0 is put on the stack of the calling context, which continues to execute normally.
+The [return data](/about) of the calling context is set as the given chunk of memory of this context.
 
 ## Stack input
 
-0. `Integer`: offset in the [memory](/about) in bytes. The starting position of the value to return.
-1. `Integer`: length in the [memory](/about) in bytes to return.
-
-## Returndata output
-
-memory[offset:offset+length]
+0. `Integer`: Offset in the [memory](/about) in bytes. The return data of the calling context.
+1. `Integer`: Length: how many bytes to copy (size of the [return data](/about).
 
 ## Examples
 
 | Memory |
 |--------|
-| `0xFF00000000000000000000000000000000000000000000000000000000000000` |
+| `0xFF01000000000000000000000000000000000000000000000000000000000000` |
 
-|  Input   | Output |
-|---------:|-------:|
-| `0x0001` | `0xFF` |
+|  Stack Input |
+|-------------:|
+|   `0x0001`   |
+
+| Calling context return data |
+|----------------------------:|
+|           `0xFF01`          |
 
 ## Error cases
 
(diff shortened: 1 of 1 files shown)


assimovt (approved):

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
 
 ## Notes
 
-Equivalent to [CREATE](/#F0), except the salt allows to create at a deterministic address.
+Abort execution, undo all state changes, return a value and refund the remaining gas to the caller.
```
peroket, line 8: Thanks, but this is not complete :-)
FD should be renamed to F5 first, and then FD can take those contents.

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+The error message returned is available to the caller in the returndata buffer.
```
peroket, line 9: Not "error message", but rather "The [return data](/about) of the calling context is set as the given chunk of memory of this context."

docs/opcodes/FD.mdx
peroket: That's nice but not complete :-)
I would say "Stop the current context execution, revert the state changes (see [STATICCALL](/#FA) for a list of state changing opcodes) and return unused gas to the caller. If the execution is stopped with REVERT, the value `0` is put on the stack of the calling context, which continues to execute normally".

"Refund" has a specific meaning, let's not use it here, it's not the same thing.

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
 
 ## Stack input
 
-0. `Integer`: value in [wei](https://www.investopedia.com/terms/w/wei.asp) to send to the new account.
-1. `Integer`: offset in the [memory](/about) in bytes, the instructions of the new account.
-2. `Integer`: how many bytes to copy (size of the instructions).
-3. `Integer`: salt, 32-byte value used to create the new account at a deterministic address.
+0. `Integer`: offset in the [memory](/about) in bytes. The starting position of the value to return.
```
peroket, line 13: The starting position of the value to return -> The return data of the calling context

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+1. `Integer`: length in the [memory](/about) in bytes to return.
```
peroket, line 14: For consistency: "how many bytes to copy (size of the [return data](/about))."

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+
+## Returndata output
```
peroket, line 16: This is not consistent, please remove that title and its content

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+
+memory[offset:offset+length]
+
+## Examples
+
+| Memory |
```
peroket, line 22: Here you can add another array (at the bottom) to show the actual bytes of the return data according to the given example memory and parameters. You can call it "Calling context return data"

peroket (changes requested):
Thanks for starting it for me :-)

docs/opcodes/FD.mdx
ArthurBraud: Sounds good, I'll change it

docs/opcodes/FD.mdx
ArthurBraud: Change it

docs/opcodes/FD.mdx
ArthurBraud: +1 changed it

ArthurBraud:
ptal @peroket

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
-2. `Integer`: how many bytes to copy (size of the instructions).
-3. `Integer`: salt, 32-byte value used to create the new account at a deterministic address.
+0. `Integer`: Offset in the [memory](/about) in bytes. The return data of the calling context.
+1. `Integer`: Length: how many bytes to copy (size of the [return data](/about).
+
+## Examples
+
+| Memory |
+|--------|
+| `0xFF01000000000000000000000000000000000000000000000000000000000000` |
```
peroket, line 20: If the zeros are not useful in your example, please remove them, to make it more clear

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+
+|  Stack Input |
+|-------------:|
+|   `0x0001`   |
```
peroket, line 24: I suppose you want to represent two values, but you did only one with extra zeros.
It should be like that:
```
| Input |
|------:|
| `0` |
| `2` |
```
You're copying two bytes in your example, so the parameter length should be 2.
Also note the title and no extra space for consistency with the rest :-)

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+
+| Calling context return data |
+|----------------------------:|
+|           `0xFF01`          |
```
peroket, line 28: Please remove the extra spaces for consistency

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+Stop the current context execution, revert the state changes (see [STATICCALL](/#FA) for a list of state changing opcodes) and return unused gas to the caller. If the execution is stopped with REVERT, the value 0 is put on the stack of the calling context, which continues to execute normally.
+The [return data](/about) of the calling context is set as the given chunk of memory of this context.
 
 ## Stack input
 
-0. `Integer`: value in [wei](https://www.investopedia.com/terms/w/wei.asp) to send to the new account.
-1. `Integer`: offset in the [memory](/about) in bytes, the instructions of the new account.
-2. `Integer`: how many bytes to copy (size of the instructions).
-3. `Integer`: salt, 32-byte value used to create the new account at a deterministic address.
+0. `Integer`: Offset in the [memory](/about) in bytes. The return data of the calling context.
```
peroket, line 13: Please remove capital letter for consistency: Offset -> offset

docs/opcodes/FD.mdx
```diff
@@ -5,18 +5,30 @@ group: System operations
...
+1. `Integer`: Length: how many bytes to copy (size of the [return data](/about).
```
peroket, line 14: Please remove "Length: " for consistency

peroket (changes requested):

peroket:
> ptal @peroket

FYI you don't need to add a comment for that, just use github "request a review" feature :-)

docs/opcodes/FD.mdx
peroket: This has not been done :-)

ArthurBraud: I removed it

peroket (approved):

Issue #67, opened 2021-11-24
Upgrade to NextJS 12


assimovt:
[NextJS 12](https://nextjs.org/blog/next-12) drastically improves refresh and build times for faster dev feedback. Also, adds support for React 18.

assimovt:
Fixed in #76

PR #76, opened 2021-12-10
feat: Add dynamic gas fee calculations

### LINKED ISSUE #57: Add more gas cost information
For each opcode with a runtime component to the gas cost, add the editable formula so that the full cost can be obtained, and the different conditions changed.

For example:
BALANCE [] warm [] cold: x cost (exclusive checkbox)
EXP 10 + 50 * [byte_len_exponent]: x cost (input field)

The constants can be obtained from ethereum-js, to stay up to date with version changes.

assimovt:
- Add documentation for dynamic gas cost and gas refunds as well as a way to estimate the dynamic gas cost
- Change "Static gas" column to "Minimum gas" to avoid misleading 0 cost
- Use Input stack variable names more consistently in the doc
- Extend opcodes.json with dynamicFee definition of user inputs
- Add a gas utility to compute fees based on inputs and Common params
- Add support for opcode specific dynamic docs with variables from Common `gasPrices` param
- Show dynamic gas fee tooltip only when it's active in the selected fork

UI:
- Add focus state on the input fields

Misc:
- Upgrade to NextJS 12 with Webpack 5
- Get rid of browserify with global ethereumjs objects
- Simplify lookup of table's opcodes object
- Fix saving of selected fork setting value
- Fix setting common chain and hardfork

Closes #57 #67 #85


components/Reference/DynamicFee.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <div key={key}>
+              <label className="block text-sm text-gray-500 mb-1" htmlFor={key}>
+                {input.label}
+              </label>
+
+              {input.type === 'number' && (
+                <Input
+                  type="number"
+                  min={0}
+                  max={1000000000000}
```
assimovt, line 86: @peroket what shall be the maximum allowed number?

components/Reference/DynamicFee.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [result, setResult] = useState('0')
+
+  // Initialize inputs with default keys & values
+  useEffect(() => {
+    const inputValues: InputValue = {}
+    Object.keys(dynamicFee?.inputs || []).map((key: string) => {
+      if (dynamicFee?.inputs[key].type === 'number') {
+        inputValues[key] = '0'
+      } else if (dynamicFee?.inputs[key].type === 'boolean') {
+        inputValues[key] = '1' // true default
```
assimovt, line 37: by default, we set booleans to 1 - True - Yes

components/Reference/DynamicFee.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const { common } = useContext(EthereumContext)
+  const [inputs, setInputs] = useState<InputValue>({})
+  const [result, setResult] = useState('0')
+
+  // Initialize inputs with default keys & values
+  useEffect(() => {
+    const inputValues: InputValue = {}
+    Object.keys(dynamicFee?.inputs || []).map((key: string) => {
+      if (dynamicFee?.inputs[key].type === 'number') {
+        inputValues[key] = '0'
```
assimovt, line 35: by default, we set numbers to 0

components/Reference/DynamicFee.tsx
peroket: I'm not sure for what you are asking, but in theory all integers can be up to the maximum value that you can encode with 256 bits (and that's immense). That default seems fine though, but we may have to raise it if we see a case where it wouldn't fit. For memory copying and even exp and the like, having such a huge number would not be really realistic though, so right now it's fine.

components/Reference/DynamicFee.tsx
peroket: I think false as default would make more sense.

components/Reference/DynamicFee.tsx
```diff
@@ -0,0 +1,119 @@
...
+      } else if (dynamicFee?.inputs[key].type === 'boolean') {
```
peroket, line 36: I don't like boolean for warm or cold (from the user perspective), can we use radio buttons?

components/Reference/index.tsx
```diff
@@ -27,11 +27,10 @@ type CustomHeaderGroup = {
   className?: string
 } & HeaderGroup<IOpcode>
 
-// FIXME: Remove once https://github.com/comitylabs/evm.codes/issues/57 is done
 const DynamicFeeTooltip = () => (
   <span
     className="inline-block pl-2 text-gray-400 dark:text-black-400"
-    data-tip="Dynamic gas portion is not included yet and coming soon."
+    data-tip="Contains dynamic gas fee, expand the row to calculate it."
```
peroket, line 33: How about something like "Has additional dynamic gas cost, expand to estimate it"?

opcodes.json
```diff
@@ -1306,6 +1328,25 @@
...
     "description": "Halt execution and register account for later deletion",
-    "dynamicFee": true
+    "dynamicFee": {
+      "inputs": {
+        "callValue": {
+          "type": "number",
+          "label": "Call value"
+        },
+        "address": {
+          "type": "boolean",
```
peroket, line 1338: Same comment here, would prefer not exposing a boolean (the only boolean we need is the one below)

util/gas.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @returns The String representation of the gas fee.
+ *
+ * Fee calculation is based on the ethereumjs-vm functions,
+ * See: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/opcodes/functions.ts
+ */
+
+export const calculateDynamicFee = (
+  opcode: IOpcode,
+  common: Common,
```
peroket, line 21: We get the fork from common as well right?

peroket (changes requested):
It's a good start, though I am missing the extra documentation for the dynamic gas cost portion (that would be also potentially fork dependent), where would it go?
I am also missing the formula displayed, did you think about that?

Lastly, I think all the inputs take a lot of space, is there a way to condense that? Like for the boolean and the warm cold, I would prefer not to ask a question, just two radio buttons warm and cold, and for the boolean maybe just a button with "empty account" as label. The documentation will explain all that needs to be explained.
Would love also that any input that is directly part of the formula be displayed directly in the formula, not taking an additional line.

Is that feasible?

components/Reference/DynamicFee.tsx
assimovt: This is the maximum number length user can input.

components/Reference/DynamicFee.tsx
assimovt: We can style it as a toggle, but its still something that has two values only, on/off - so its a boolean in the opcode.json meta

util/gas.ts
assimovt: Yes, its the same common context that's used in the reference table. So, you can get the hardfork from it

components/Reference/DynamicFee.tsx
peroket: Yes in the json it's fine, would like to change how it's shown :-)
A button that toggles with written warm on side and cold the other is perfect, like @dorlevi suggested

components/Reference/DynamicFee.tsx
assimovt: > It's a good start, though I am missing the extra documentation for the dynamic gas cost portion (that would be also potentially fork dependent), where would it go? I am also missing the formula displayed, did you think about that?
> 
> Lastly, I think all the inputs take a lot of space, is there a way to condense that? Like for the boolean and the warm cold, I would prefer not to ask a question, just two radio buttons warm and cold, and for the boolean maybe just a button with "empty account" as label. The documentation will explain all that needs to be explained. Would love also that any input that is directly part of the formula be displayed directly in the formula, not taking an additional line. And that way the result gas cost would be the last term of the formula, right now I find it counter intuitive that it be in the title.
> 
> Is that feasible?

Re "missing dynamic gas cost" - we have a handler to the `common` object, so everything can be picked up from there.

Re "formulas" - Some formulas are not single-line that you can easily display, isn't it? So, I have a hard time imagining how it would look in the UI. As you remember, we first tried to do inline-editing of the formulas, but it didn't work for all opcodes.

peroket: > missing dynamic gas cost
I have to say I do not understand your answer. I am asking where do I write the documentation that I want to add for each opcode with dynamic gas cost. Markdown description. Not something we get from the code.

> formulas
Yes for the formulas not all can, but can we make the ones that can inlined with their parameters? Or is it too much effort? (Remember I am asking here not as a technical person, if it's too hard to do for some technical reason you know best and you can say so)

> condensed representation
An idea could be to put two variables per line also, for example, the numbers do not need all that horizontal space.

assimovt: > missing dynamic gas cost

Aga, gotcha. So, if it's a simple line of text, we can add it in the opcode.json as description metadata and show it in the UI near each input. If it's more detailed and needs markdown, then we add it to the opcode markdown files, right in the top as usual, with some heading.

> formulas

Of course, technically it's feasible, but the question is if it makes sense UI/UX-wise, worth the implementation effort, and code maintainability. If we were to go that route, then each formula will have to be displayed differently in the UI, some will be a single line of editable text, another multi-line inputs with radios, etc. This makes the UI inconsistent and code more brittle, as now you have to define and build it differently. With the current approach, it's all driven by a single metadata file and the UI is auto-generated for any kind of opcode (regardless of how many formula terms there are), in a consistent manner. It's simple and serves the purpose. So, IMHO, it's not worth the effort to show a few opcodes as formulas with inline-editing capability.

peroket: > missing dynamic gas cost

It's unlikely to be a simple line of text, and it does not fit in the opcodes markdown files, as we need it to be versioned. Though I guess if you can add a filter in the markdown so that we can write all descriptions but you only display the one from the current fork, that sounds nice to me.

> formulas

That's fair, let's keep it that way then. Let's just adapt how we display the inputs. For example, instead of a line for name and then a line for the value, can we try to put them both on the same line (just thinking that if that takes too much vertical space, it's kind of hiding what's below). I am also wondering what would be the better placement. I think maybe it would be better that the dynamic gas cost be the last of the extra info, that way it's less problematic when it takes so much vertical space, and the description of the opcode should probably be first anyway.

To display the formulas, would be awesome if I could write them in the markdown too. If it's versioned by fork, I guess I can hardcode the constant costs and we don't need to read them dynamically from ethereum js. Though I guess it would still be convenient if nothing changed but the constant values, would not have to copy paste everything but the constants. And would still need a placeholder to write the result of the formula (the title is not the best place).

Would be nice in that case also to have a way to have a "default", or maybe even better, the filter would take the oldest hardfork it can find. That way, if it did not change, I don't have to copy paste for each version.

I am asking a lot, sorry :-)
Is any of those realistic?

assimovt: So after some discussion and thoughts with @peroket, here is what we will do to improve doc experience around dynamic gas:

1. Move dynamic gas calculation form to the side (floating), and on narrower screens place at the bottom of the doc.
2. Add support for dynamic variables in the opcode markdown documents, starting with EthereumJS Common params. We will namespace them with `eths` and wrap in curly braces, for instace `expByte` from the library would be `{ethjs|expByte}` in the doc.
3. We will start with the same documents, but if it appears that there is a substantial difference in gas calculation formulas & documentation between the forks, we will add doc overrides, for example: `0a/london.mdx`

opcodes.json
```diff
@@ -272,7 +272,9 @@
     "input": "address",
     "output": "balance",
     "description": "Get balance of the given account",
-    "dynamicFee": true
+    "dynamicFee": {
+      "since": "berlin"
```
assimovt, line 276: @peroket wit this commit, you can add `since` (inclusive) and the dynamic part will only be visible if the selected fork is past the since

pages/about.tsx
```diff
@@ -121,27 +123,170 @@ const AboutPage = () => {
...
+        </a>
+        .
       </p>
+
+      <H3 className="mb-4">Intrinsic Gas</H3>
+      <p className="pb-8">
+        Each transaction has an intrinsic cost of 21000 gas. Creating a contract
+        costs 32000 gas, on top of the transaction cost. And finally, the
+        calldata costs 4 gas per byte equal to 0, and 16 gas for the others (64
+        before Istanbul fork). This cost is payed from the transaction before
```
assimovt, line 178: @peroket nit: consistency: paid vs payed

pages/about.tsx
```diff
@@ -121,27 +123,170 @@ const AboutPage = () => {
...
+            <br />
+            memory_cost = (memory_size_word ** 2) / 512 + (3 * memory_size_word)
+          </code>
+        </Pre>
+      </p>
+
+      <p className="pb-4">
+        <p className="pb-4">
+          When a memory expansion is triggered however, only the additional
+          chunk of memory has to be payed. The cost of memory expansion for a
```
assimovt, line 208: same here

pages/about.tsx
```diff
@@ -121,27 +123,170 @@ const AboutPage = () => {
...
+          </li>
+        </ul>
+        When a context reverts, the sets are also reverted to the state they had
+        before that context.
+      </p>
+
+      <H3 className="mb-4">Gas refunds</H3>
+      <p className="pb-8">
+        Some opcodes can trigger gas refunds, which reduces the gas cost of a
+        transaction. Coming soon.
```
assimovt, line 275: let's remove 'Coming soon'

assimovt (commented):
@peroket added a couple of minor comments about the texts

pages/about.tsx
peroket: Thanks, fixed

pages/about.tsx
peroket: Fixed

pages/about.tsx
peroket: Why?

peroket (approved):

pages/about.tsx
assimovt: Maybe better to link to an issue, that's more actionable - you could subscribe, etc.

peroket: I did the refund, so no longer relevant

pages/about.tsx
```diff
@@ -121,27 +123,175 @@ const AboutPage = () => {
...
+            <OLink opcode="F1" title="CALL" />,{' '}
+            <OLink opcode="F2" title="CALLCODE" />,{' '}
+            <OLink opcode="F4" title="DELEGATECALL" />,{' '}
+            <OLink opcode="FA" title="STATICCALL" />,{' '}
+            <OLink opcode="F0" title="CREATE" />,{' '}
+            <OLink opcode="F5" title="CREATE2" /> and{' '}
+            <OLink opcode="FF" title="SELFDESTRUCT" />.
+          </li>
+          <li className="ml-6">
+            Slots: a set of contract address and their slot keys that have been
```
ArthurBraud, line 259: Maybe precise `storage slot`?

pages/about.tsx
```diff
@@ -121,27 +123,175 @@ const AboutPage = () => {
...
+        ). Each opcode that can is mentionned in the <OLink title="reference" />
+        . Note also that an opcode with a byte size parameter of 0 will not
+        trigger a memory expansion, regardless of its offset parameters.
+      </p>
+
+      <H3 className="mb-4">Access sets</H3>
+      <p className="pb-8">
+        Access sets have been introduced in the hardfork <b>Berlin</b>. They are
+        kept per transaction (and not per call context). Two of them exist: the
+        touched contracts addresses, and the touched contract slots. If an
```
ArthurBraud, line 234: Maybe precise `storage slot` ?

docs/opcodes/31/berlin.mdx
```diff
@@ -0,0 +1,3 @@
+## Gas
+
+The static cost is {gasPrices|balance}. If the accessed address is warm, the dynamic cost is {gasPrices|warmstorageread}. Otherwise the dynamic cost is {gasPrices|coldaccountaccess}. See section [access sets](/about).
```
ArthurBraud, line 3: Very nit: Maybe we could add the formular here like gas = n cold accesses * {gasPrices|coldaccountaccess} + n warm accesses * {gasPrices|warmstorageread}

docs/opcodes/37/homestead.mdx
```diff
@@ -0,0 +1,8 @@
+## Gas
```
ArthurBraud, line 1: I think the group should be `System operations` here?

pages/about.tsx
peroket: good point, done

pages/about.tsx
peroket: also done

docs/opcodes/37/homestead.mdx
peroket: This is the groups from the yellow paper, they are correct :-)

Also fyi, the groups are set in `docs/opcodes/37.mdx`, not this file.

peroket (approved):

dorlevi:
glad to see this out 👏

PR #92, opened 2021-12-30
feat: Add precompiled contracts reference

peroket:


components/Editor/Header.tsx
```diff
@@ -66,6 +66,13 @@ const EditorHeader = ({
             isChecked={codeType === CodeType.Bytecode}
             onChange={onCodeTypeChange}
           />
+
+          <Radio
+            text="Mnemonic"
+            value={CodeType.Mnemonic}
+            isChecked={codeType === CodeType.Mnemonic}
+            onChange={onCodeTypeChange}
+          />
```
assimovt: This does not fit, so I've moved it to the Select box

components/Editor/index.tsx
```diff
@@ -146,8 +148,75 @@ const Editor = ({ readOnly = false }: Props) => {
...
+            return opcode.name === line
+          })
+          if (typeof code === 'undefined') {
+            log('Unknown mnemonic: ' + line, 'warn')
+            return
+          }
+
+          bytecode += code.code
+        }
+      }
```
assimovt, line 215: Would be nice to extract it into util, for example in `util/compiler.ts`, which would run these and throw an error. Then this component could catch it and log to UI.

assimovt (changes requested):
I refactored a bunch of duplication in here in ee367a14cd0ef8f3a631ddf348ea291c2fd9637c.  There is just one function extraction that I think you could take over.

pages/precompiled.tsx
```diff
@@ -0,0 +1,137 @@
...
+      </section>
+    </>
+  )
+}
+
+PrecompiledPage.getLayout = function getLayout(page: NextPage) {
+  return <HomeLayout>{page}</HomeLayout>
+}
+
+export const getStaticProps = async () => {
```
assimovt, line 85: This one is an exact copy of the index.tsx, except the docs path. I tried moving it into a separate function, but it breaks the server-side webpack build. I will keep looking for a solution to DRY this up.

pages/precompiled.tsx
```diff
@@ -1,41 +1,31 @@
...
-  <Link href={link ? `/${link}` : '/'} passHref>
-    <a className="underline font-mono">{title}</a>
-  </Link>
-)
+import generateDocs from './static/generateDocs'
 
 // It seems the memory expansion computation and constants did not change since frontier, but we have to keep an eye on new fork to keep this up to date
 const PrecompiledPage = ({
-  precompiledDocs,
+  opcodeDocs,
```
peroket, line 23: Calling it opcode kinda triggers me. Shall we just call it "docs"? And same for other places, not use opcodes when it may also be a precompiled

pages/precompiled.tsx
```diff
@@ -1,41 +1,31 @@
...
   gasDocs,
 }: {
-  precompiledDocs: IPrecompiledDocs
-  gasDocs: IPrecompiledGasDocs
+  opcodeDocs: IOpcodeDocs
```
peroket, line 26: Like here the type should be changed also, "IDocs"?

types/index.ts
```diff
@@ -44,13 +25,7 @@ export interface IPrecompiled {
...
 
 export interface IOpcodeMeta {
-  input: string
-  output: string
-  description: string
-}
-
-export interface IPrecompiledMeta {
-  name: string
+  name?: string
```
peroket, line 28: I think it would help a lot if we added comments on when we expect this to be filled or not

util/gas.ts
```diff
@@ -401,12 +401,12 @@ function multComplexityEIP2565(x: BN): BN {
...
  */
 export const calculatePrecompiledDynamicFee = (
-  precompile: IPrecompiled,
+  contract: IOpcode,
   common: Common,
   inputs: any,
 ) => {
   let result = null
-  switch (precompile.address) {
+  switch (contract.code) {
```
peroket, line 409: contract.id?

lib/useActions.tsx
```diff
@@ -18,6 +18,16 @@ const useActions = () => {
       subtitle: 'Opcodes reference',
       icon: <Icon name="home-2-line" />,
     },
+    {
+      id: 'precompiled',
+      name: 'Precompiled',
+      shortcut: ['a'],
```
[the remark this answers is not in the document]
peroket: p?

components/Editor/index.tsx
peroket: I was hoping you could rewrite it, because I assume it is not coded in the best typescript way?

util/gas.ts
peroket: Or maybe even better, contract.nameOrAddress?

lib/useActions.tsx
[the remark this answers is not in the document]
peroket: or 'c'?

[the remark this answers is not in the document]
peroket: Otherwise we could just use numbers, 1, 2... in the order of the links above, like it works for chrome and windows and more

assimovt (approved):
👏🏽

PR #94, opened 2022-01-12
feat: Add value and calldata fields for playground

assimovt:


peroket (approved):
Approved by @assimovt and myself

PR #96, opened 2022-01-13
feat: reflect code, callvalue and calldata in url to be able to share…

peroket:


assimovt (commented):
I think we could make it better. 

1. I don't think we should populate the URL history on every code change, and instead have a button somewhere to get a permalink to the clipboard, which could then be shared with people. You would still need to capture the values in `useEffect` ofc.
 
There are a couple of reasons why I think it's not a good approach: a) every slightest code change will hit the browser history and it's not needed as the user won't probably even want to share the URL b) it makes the URL unnecessarily bloated. 

2. I am not keen on `@kunigi/string-compression`. It seems like it's not used much (at all?) and the maintainer site is nonexistent. No activity either. Additionally, the encoded strings it produces are super ugly and do not really compress much. I would rather use plain text query params for the code type/call value/etc (like code=bytecode&value=XXX), which will make it more readable, and for the code value, itself use Base64 encoded string.

What do you think?

peroket:
> I think we could make it better.
> 
> 1. I don't think we should populate the URL history on every code change, and instead have a button somewhere to get a permalink to the clipboard, which could then be shared with people. You would still need to capture the values in `useEffect` ofc.
> 
> There are a couple of reasons why I think it's not a good approach: a) every slightest code change will hit the browser history and it's not needed as the user won't probably even want to share the URL b) it makes the URL unnecessarily bloated.
> 
> 2. I am not keen on `@kunigi/string-compression`. It seems like it's not used much (at all?) and the maintainer site is nonexistent. No activity either. Additionally, the encoded strings it produces are super ugly and do not really compress much. I would rather use plain text query params for the code type/call value/etc (like code=bytecode&value=XXX), which will make it more readable, and for the code value, itself use Base64 encoded string.
> 
> What do you think?

1. I was thinking of doing something like that, but then I don't have a good idea of where to place such a button, and I like the simplicity of no buttons, but if you have an idea for the UI sure.
2. I don't need the string to be readable. I do like that now it's visible for the value and calldata, but the code itself I prefer to encode it because I want to let people have more freedom about what they can share (which is targeted mostly for Mnemonic code). And that encoding works really well for that (in my test with our example solidity it saved ~200 characters compared to plain text, and with mnemonic basically only the comments are not repeated), so I will be hard to convince to change it :-)

assimovt:
@peroket So I've added UI for sharing permalinks via clipboard in #944764c (along with mobile responsiveness for that row buttons).

dorlevi:
> I think we could make it better.
> 
> 1. I don't think we should populate the URL history on every code change, and instead have a button somewhere to get a permalink to the clipboard, which could then be shared with people. You would still need to capture the values in `useEffect` ofc.
> 
> There are a couple of reasons why I think it's not a good approach: a) every slightest code change will hit the browser history and it's not needed as the user won't probably even want to share the URL b) it makes the URL unnecessarily bloated.
> 
> 2. I am not keen on `@kunigi/string-compression`. It seems like it's not used much (at all?) and the maintainer site is nonexistent. No activity either. Additionally, the encoded strings it produces are super ugly and do not really compress much. I would rather use plain text query params for the code type/call value/etc (like code=bytecode&value=XXX), which will make it more readable, and for the code value, itself use Base64 encoded string.
> 
> What do you think?

I know you already resolved it but just wanted to thank Tair for his feedback on 1. I think it's spot on and I appreciate it. I do  understand and agree with @peroket rationale on encoding/compressing

components/Editor/index.tsx
```diff
@@ -82,49 +93,94 @@ const Editor = ({ readOnly = false }: Props) => {
   const [callValue, setCallValue] = useState('')
   const [unit, setUnit] = useState(ValueUnit.Wei as string)
 
-  const handleWorkerMessage = (event: MessageEvent) => {
-    const { code: byteCode, warning, error } = event.data
+  const log = useCallback(
+    (line: string, type = 'info') => {
+      // See https://blog.logrocket.com/a-guide-to-usestate-in-react-ecb9952e406c/
```
assimovt, line 98: lets remove this comment

components/Editor/index.tsx
```diff
@@ -82,49 +93,94 @@ const Editor = ({ readOnly = false }: Props) => {
...
+      setOutput((previous) => {
+        const cloned = previous.map((x) => ({ ...x }))
+        cloned.push({ type, message: line })
+        return cloned
```
assimovt, line 102: you could use an array spreading here instead of a func, smth like:

```
setOutput([...output, { type, message: line }])
```

components/Editor/index.tsx
```diff
@@ -82,49 +93,94 @@ const Editor = ({ readOnly = false }: Props) => {
...
+    } else {
+      const initialCodeType: CodeType =
+        getSetting(Setting.EditorCodeType) || CodeType.Yul
+
+      setCodeType(initialCodeType)
+      setCode(examples[initialCodeType][0])
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [settingsLoaded])
+  }, [settingsLoaded && router.isReady])
```
assimovt, line 183: I think these should be comma-separated, so it tracks both separately?

components/Editor/index.tsx
```diff
@@ -358,13 +419,21 @@ const Editor = ({ readOnly = false }: Props) => {
                   classNamePrefix="select"
                   menuPlacement="auto"
                 />
+
+                <Button
+                  onClick={handleCopyPermalink}
+                  transparent
+                  padded={false}
```
assimovt, line 426: I'd add a tooltip here to make it clear what it does, just pass in tooltip props as in the debugging step button icons

assimovt (changes requested):
Some minor tweaks requested

components/Editor/index.tsx
peroket: In the link they say you have to make sure to do a deep copy, including the object in the array, and as far as I understood your example would not do that

components/Editor/index.tsx
peroket: I want that function to be called when both the settings are loaded and the router is ready, it's not the way to do it?

assimovt: Ah in that case it's ok. I thought you wanted to run it when any change on deps happen. 👌🏽

components/Editor/index.tsx
peroket: Good idea, done

components/Editor/index.tsx
assimovt: Oh, you're right. We're having an array of objects, and not primitives here. That's cool then

components/Editor/index.tsx
peroket: I would rather keep it, it explains why the code looks weird :-)
And this is something that's rather unexpected, so it's good to keep the link

assimovt (approved):
:shipit:

components/Editor/index.tsx
peroket: We agreed to keep it

PR #103, opened 2022-01-19
fix: get url from vercel environment

peroket:


dorlevi (approved):
works for me, I am still curious if there is a way to fix the original version as it seemed "cleaner"  but we can ask @assimovt at some point

PR #105, opened 2022-01-20
fix: getAbsoluteURL and remove unnecessary dot files

assimovt:
The dotfiles introduced in #103 and #104 are not needed since `NEXT_PUBLIC_VERCEL_URL` is already one of the [exposed environment variable](https://vercel.com/docs/concepts/projects/environment-variables) by Vercel. The original bug was not using the correct prefix in the `getAbsoluteURL` of `util/browser.ts`, which is fixed in this PR. After deploying this one, the custom environment variable in the Vercel settings is no longer required too - check out the Preview environment where it works w/o it.

### DIFF
--- a/util/browser.ts
+++ b/util/browser.ts
@@ -4,8 +4,8 @@ export const isMac =
     : false
 
 export const getAbsoluteURL = (path = '') => {
-  const baseURL = process.env.VERCEL_URL
-    ? `https://${process.env.VERCEL_URL}`
+  const baseURL = process.env.NEXT_PUBLIC_VERCEL_URL
+    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
     : 'http://localhost:3000'
   return baseURL + path
 }
--- a/.env.development
+++ b/.env.development
@@ -1,2 +1,0 @@
-# .env.development
-NEXT_PUBLIC_VERCEL_URL=http://localhost:3000
--- a/.env.production
+++ b/.env.production
@@ -1,2 +1,0 @@
-# .env.production
-NEXT_PUBLIC_VERCEL_URL=https://${VERCEL_URL}
--- a/components/Editor/index.tsx
+++ b/components/Editor/index.tsx
@@ -20,6 +20,7 @@ import SCEditor from 'react-simple-code-editor'
 import { EthereumContext } from 'context/ethereumContext'
 import { SettingsContext, Setting } from 'context/settingsContext'
 
+import { getAbsoluteURL } from 'util/browser'
 import {
   getTargetEvmVersion,
   compilerSemVer,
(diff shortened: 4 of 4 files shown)


peroket:
Just as a question: are you sure we don't need that variable setup in vercel? Because when it was not there, using the exact same variable as you, the link was giving the link of the deployed address (so something like that `https://evm-codes-git-fix-absolute-url-util-comity.vercel.app`), and not the link of the "official" website.
So I think the variable should not be removed.

peroket (approved):

util/browser.ts
```diff
@@ -4,8 +4,8 @@ export const isMac =
     : false
 
 export const getAbsoluteURL = (path = '') => {
-  const baseURL = process.env.VERCEL_URL
-    ? `https://${process.env.VERCEL_URL}`
+  const baseURL = process.env.NEXT_PUBLIC_VERCEL_URL
```
peroket, line 7: Just for my own curiosity: we are doing the same thing, why is your way better?

assimovt:
> Just as a question: are you sure we don't need that variable setup in vercel? Because when it was not there, using the exact same variable as you, the link was giving the link of the deployed address (so something like that `https://evm-codes-git-fix-absolute-url-util-comity.vercel.app`), and not the link of the "official" website. So I think the variable should not be removed. @assimovt

That's from the preview environment, right? We still want to have those links in the branch previews. I think in production (main) it should be resolved to the custom domain - but we can easily verify it once this one is deployed, and keep it if necessary.

util/browser.ts
assimovt: Few reasons:

1. We dont need .env.development just yet as we only have one env-specific configuration, so we can just default to it here.
2. If we ever change how the URL is constructed or the variable name, we will only need to modify getAbsoluteURL util, instead of going to every place it was used.
3. The .env.production is also redundant as that's the environment variable automatically exposed by Vercel.

So, this actually fixes the original bug that was there, instead of working around it with dotfiles.

peroket:
> > Just as a question: are you sure we don't need that variable setup in vercel? Because when it was not there, using the exact same variable as you, the link was giving the link of the deployed address (so something like that `https://evm-codes-git-fix-absolute-url-util-comity.vercel.app`), and not the link of the "official" website. So I think the variable should not be removed. @assimovt
> 
> That's from the preview environment, right? We still want to have those links in the branch previews. I think in production (main) it should be resolved to the custom domain - but we can easily verify it once this one is deployed, and keep it if necessary.

The variable in vercel is for the www.evm.codes for the real website, the PUBLIC URL thingy works as is for the preview

util/browser.ts
peroket: No there was two bugs: the wrong variable used (which is also fixed here), and the wrong setup setup in vercel (which has been fixed in our settings).
The production env is not necessary indeed, but the development one is just another way to do what you are doing in your function to make it work locally.
But fine with me, just was wondering if there was something intrinsically better.

PR #111, opened 2022-01-31
feat: add gas refund estimate for 55, and fix gas refund formula

peroket:


components/Reference/DynamicFee.tsx
```diff
@@ -27,12 +27,20 @@ const DynamicFee = ({ referenceItem, fork }: Props) => {
 
   const { common } = useContext(EthereumContext)
   const [inputs, setInputs] = useState<InputValue | undefined>()
-  const [result, setResult] = useState('0')
+  const [gas_cost, setGasCost] = useState('0')
+  const [gas_refund, setGasRefund] = useState('0')
+  const [can_refund, setCanRefund] = useState(false)
```
assimovt, line 32: nit: camelCase

components/Reference/DynamicFee.tsx
```diff
@@ -27,12 +27,20 @@ const DynamicFee = ({ referenceItem, fork }: Props) => {
...
     if (common) {
       try {
-        setResult(calculateDynamicFee(referenceItem, common, inputs))
+        setGasCost(calculateDynamicFee(referenceItem, common, inputs))
+
+        const refund = calculateDynamicRefund(referenceItem, common, inputs)
+        if (refund != null) {
+          setCanRefund(true)
+          setGasRefund(refund)
+        }
```
assimovt, line 43: should there be a fallback to reset back to 0/false?

components/ui/vector/Remix.tsx
```diff
@@ -68,6 +68,9 @@ function SvgRemix(props: any) {
         <symbol id="remix_svg__ri-play-circle-line" viewBox="0 0 32 32">
           <path d="M16 29.333C8.636 29.333 2.667 23.364 2.667 16S8.636 2.667 16 2.667 29.333 8.636 29.333 16 23.364 29.333 16 29.333zm0-2.666c5.891 0 10.667-4.776 10.667-10.667S21.891 5.333 16 5.333 5.333 10.109 5.333 16 10.109 26.667 16 26.667zM14.163 11.22l6.505 4.336a.535.535 0 01.002.887l-.002.001-6.507 4.336a.533.533 0 01-.828-.442v-8.675a.533.533 0 01.831-.441l-.002-.001z" />
         </symbol>
+        <symbol id="remix_svg__ri-reply-fill" viewBox="0 0 32 32">
+          <path d="M11 20L1 12l10-8v5c5.523 0 10 4.477 10 10 0 .273-.01.543-.032.81C19.46 16.95 16.458 15 13 15h-2v5z" />
+        </symbol>
```
assimovt, line 73: we have an icons metadata file in `vector/icons.remixicon` that contains the full list of icons we use in the app. This helps us to update `Remix.tsx` automatically, see `vector/README.md`. So, either update it manually or import and export it in Remixicon (recommended, less error prone).

util/gas.ts
```diff
@@ -85,7 +85,11 @@ function sstoreCost(common: Common, inputs: any): BN {
     }
   } else if (common.gteHardfork('istanbul')) {
     if (inputs.newValue === inputs.currentValue) {
-      return new BN(common.param('gasPrices', 'sstoreNoopGasEIP2200'))
+      if (common.gteHardfork('berlin') && inputs.cold !== '1') {
+        return new BN(common.param('gasPrices', 'warmstorageread'))
+      } else {
+        return new BN(common.param('gasPrices', 'sstoreNoopGasEIP2200'))
+      }
```
assimovt, line 92: nit: unnecessary else (similarly below)

assimovt (changes requested):
some minor things, but otherwise looks good

components/Reference/DynamicFee.tsx
peroket: done

components/Reference/DynamicFee.tsx
peroket: It seems to have the wanted behaviour already, what would it change?

util/gas.ts
peroket: I don't see it? "sstoreNoopGasEIP2200" is not anywhere else, and if I don't return here it will never reach a "return" statement either.

peroket: Or do you mean I can just remove the "else" keyword? That's a style preference, and personnaly I prefer it this way, it's more clear what's happening to me.

components/ui/vector/Remix.tsx
peroket: I missed that. I copied the path from remixicon directly, still seems simpler to me :-)
Added the new icon in the export file though.

components/Reference/DynamicFee.tsx
assimovt: I haven't tested, but I guess if refund ever returns null and the component has previous state set, it will not reset back to null. So, this ensures it never happens, and is a good precaution for resettings the state.

util/gas.ts
assimovt: Yeah, it's just a js [styling convention](https://eslint.org/docs/rules/no-else-return), which we don't have in Eslint probably. I will let you decide if we want it, but I personally find it redundant.

assimovt (approved):

components/Reference/DynamicFee.tsx
peroket: Done

util/gas.ts
peroket: Yep I prefer it that way, so I won't touch it.

assimovt (approved):

PR #113, opened 2022-02-15
feat: fix reference link for id 0 and precompiled, and add permalink for each reference line

peroket:


assimovt (approved):
ui nit: I think it'd look nicer if it was after the opcode number.

PR #119, opened 2022-03-07
fix: add stack output to CREATE documentation

OliverNChalk:
I'm assuming the stack output of CREATE was accidentally omitted, rather than intentionally (given it is included in CREATE2 and most opcodes document their stack output). Address is correctly mentioned in the "STACK OUTPUT" portion of the opcode table, so this is just a minor issue with aligning the docs


peroket (approved):
Thank you for your help :-)

PR #120, opened 2022-03-07
About changes

dorlevi:
adding EVM resources

### DIFF
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -281,6 +281,7 @@ const AboutPage = () => {
         </a>{' '}
         for the cost descriptions.
       </em>
+
       <H2 className="mb-4">Other EVM Related Resources</H2>
       <p className="pb-8">
         <a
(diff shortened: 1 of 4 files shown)


pages/about.tsx
```diff
@@ -266,21 +266,68 @@ const AboutPage = () => {
         <b>London</b>, otherwise to a fifth. Starting from the hardfork{' '}
         <b>London</b> also, only <Link to="#55" title="SSTORE" /> may trigger
         refunds. Before that, <Link to="#FF" title="SELFDESTRUCT" /> could also
-        trigger refunds.
+        trigger refunds. \
```
peroket, line 269: You have an extra '\' to remove

pages/about.tsx
```diff
@@ -266,21 +266,68 @@ const AboutPage = () => {
...
+        <em>
```
peroket, line 270: Can you put it in a different paragraph? (in its own `<p></p>`)

peroket (changes requested):
Thanks for the links. Looks good, but please revert your local config changes from the state.
Basically it should only contain the changes to the 'pages/about.tsx' file.

PR #127, opened 2022-04-07
Disable 'cold' option when estimating SSTORE gas costs with known warm storage

### LINKED ISSUE #118: Ambiguity in storage gas costs
image

In the above case, the 'key' can never be cold. So the option for that should be disabled.

malexw:
When estimating the gas cost of the SSTORE instruction, we provide an option to tell the calculator if the storage is warm or cold. However, if the current value of the slot is different from the original value, that implies that the slot has already been touched as part of this transaction and therefore can't be cold.

This change disables the 'cold' option when current value != original value and visually causes the option to appear as though the user has selected 'No' (meaning 'not cold'), when disabled. If current value and original value are later changed to be equal again, the 'cold' option is restored to whatever value was selected before.

Resolves #118

### DIFF
--- a/components/Reference/DynamicFee.tsx
+++ b/components/Reference/DynamicFee.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useState, useContext } from 'react'
+import Common from '@ethereumjs/common'
 
 import debounce from 'lodash.debounce'
 import { IReferenceItem } from 'types'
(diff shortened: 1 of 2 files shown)


peroket (approved):
I don't understand the code changes too much, but it works, so looks good to me :-)

assimovt (approved):
LGTM

PR #139, opened 2022-05-17
prepend 0x to opcodes codes

berdim99:
The opcode numbers are in hex, but the UI doesn't display them as such. This fixes that.

Before:
Screen Shot 2022-05-17 at 4 17 14 PM

After:

Screen Shot 2022-05-17 at 4 17 05 PM

### DIFF
--- a/components/Reference/columns.tsx
+++ b/components/Reference/columns.tsx
@@ -11,7 +11,7 @@ const columns = (isPrecompiled: boolean) => [
   {
     Header: !isPrecompiled ? 'Opcode' : 'Address',
     accessor: 'opcodeOrAddress',
-    className: !isPrecompiled ? 'uppercase' : undefined,
+    className: undefined,
     filter,
     width: 48,
   },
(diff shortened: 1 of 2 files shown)


berdim99:
(this is not ready yet - broke seeing details of opcodes :( )

peroket (commented):
Thank you for proposing that :-)

However this changes only affects what is displayed, but not the search. If you try and search opcode `0x03` for example it will not work.
Same for the link, linking to an opcode is done without the '0x'.

Because of that I'm not sure it's a good idea to change the representation in the reference as well.

berdim99:
I see. That's fair. One thing that we didn't want to change is to make sure that existing URLs will still work.
I'd argue that this change is still good, as the number "55" means nothing without the "0x" prefix, but I can look into improving it -- ideally: 1) make searching for 0x55 work, and, 2) ideally, make new links have 0x prefix, while having existing links still work.

peroket:
> I see. That's fair. One thing that we didn't want to change is to make sure that existing URLs will still work. I'd argue that this change is still good, as the number "55" means nothing without the "0x" prefix, but I can look into improving it -- ideally: 1) make searching for 0x55 work, and, 2) ideally, make new links have 0x prefix, while having existing links still work.

Yes if you do it that way that would be fine.

berdim99:
closing this PR as I'm not planning to get back to it anytime soon

PR #144, opened 2022-06-10
Section now are "anchorable"

StErMi:
I needed to reference a single section of the page, so I thought to make them anchorable.
Things work, but I run into a couple of issues:

- when you click on an anchor, it seems that the page scroll too much. Probably it's because of the sticky header?
- the UI (having each anchor always visible, maybe it's not the best?) could be improved. Perhaps make the anchor icon visible only when user hover over it?

Looking forward for feedback

### DIFF
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -5,7 +5,6 @@ import HomeLayout from 'components/layouts/Home'
 import { Container, H1, H2, H3, Icon, RelativeLink } from 'components/ui'
 import { Pre } from 'components/ui/Doc'
 
-
 type SectionWrapperProps = {
   header: React.ReactNode
   anchorKey: string
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -18,7 +18,10 @@ const SectionWrapper: React.FC<SectionWrapperProps> = ({
 }) => {
   return (
     <>
-      <div id={anchorKey} className="font-mono mb-4 justify-start relative items-center scroll-mt-14">
+      <div
+        id={anchorKey}
+        className="font-mono mb-4 justify-start relative items-center scroll-mt-14"
+      >
         <Link href={`/about#${anchorKey}`}>
           <a className="absolute -left-6">
             <Icon name="links-line" className="text-indigo-500" />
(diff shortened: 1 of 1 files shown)


malexw:
Thanks for the PR!

I think you're correct about the issue with the scroll behavior and sticky header. This might be something that can be solved with `scroll-margin` or `scroll-padding` CSS. We've just merged an upgrade to Tailwind 3 that provides helper classes for these properties, so that might be something to play around with. Here are the docs if you'd like more info: https://tailwindcss.com/docs/scroll-padding

I definitely agree with you that only showing the anchor icon on hover would clean things up a bit. The [lodash docs](https://lodash.com/docs/4.17.15) are a good example of this.

StErMi:
> Thanks for the PR!
> 
> I think you're correct about the issue with the scroll behavior and sticky header. This might be something that can be solved with `scroll-margin` or `scroll-padding` CSS. We've just merged an upgrade to Tailwind 3 that provides helper classes for these properties, so that might be something to play around with. Here are the docs if you'd like more info: [tailwindcss.com/docs/scroll-padding](https://tailwindcss.com/docs/scroll-padding)
> 
> I definitely agree with you that only showing the anchor icon on hover would clean things up a bit. The [lodash docs](https://lodash.com/docs/4.17.15) are a good example of this.

I think that's easily doable also with only tailwind to display the icon only on on-over. I'll make some test tomorrow.
The scroll issue won't be easy to fix I'm afraid but I'll try my best tomorrow

StErMi:
I have updated the code to support relative-absolute positions of the anchor in style with the Ethereum blog.
I've tried to handle the scroll position but it seems that it does not work with your suggestion :(

malexw (approved):
Thanks again for the PR! I think the style for anchor links that you picked looks good.

I think that the `scroll-margin` changes probably weren't working for you because the update to Tailwind 3 was merged to `main` after this branch was created so I just had to pull it in. To clean up the scroll positioning, I moved the anchor ID to be on the same div as the section title and added a `scroll-margin-top: 14rem` property to the div.

StErMi:
@malexw the final result looks really nice!

PR #148, opened 2022-06-16
Feature add solidity method run

jonahcui:
I need more args to run my contract's method or deploy my contract. 
*  I add a solidity method selector component to implement method selection in solidity. 
*  I add constructor info when compiling the contract.  so that we can add more args in deploying the contract.

Other features or Bug fixes:
* Fix call value not work in deploying the contract. Because we need to update the worker's ```onmessage``` method when the call value changed.
* Add opcode id in opcode window. when the opcode is ```jump``` or ```jump```, we need to know where the code will run in the next step.

Some Component Preview:
CleanShot 2022-06-16 at 16 39 38@2x


dorlevi:
Thank you for contributing we will look into this PR shortly!

peroket (commented):
Taking from a user point of view:
Thanks for adding those changes, they are great additions, but I don't think they are complete. First, having to use `run` first before you can get the correct list of constructor arguments, meaning you have to launch a failing transaction, is not ideal.
Second, what format should the arguments be written in? What types are supported? This is missing.
The run method also is not working correctly. I tried to use it several times, and it worked only the first time. The opcode window is not reset, I can not play the transaction. When I set a parameter, it keeps it even if I call a function without parameters afterwards.

And finally, there is also the question, how are several contracts handled?

The opcode IDs are a great idea also, maybe  they should be moved to their own PR to be merged sooner?

So in conclusion, thank you for your effort, it's definitely showing nice new functionalities, but it is not ready to be deployed as it is. I believe adding that kind of functionalities needs a bigger rewrite and efforts, and more prfound UI changes. Happy to review it again if you can spare the time, though it would be nice to make sure we are in agreement with the changes before you spend too much time on them :-)

jonahcui:
> peroket

Yes, I think your point is right. I think we may need more discussion to solve these problems.
* The ```run``` need to be split into two methods, ```compiler`` and ```run```. 
* The arguments format is ```args1,args2,args3```, this format is the same as the remix.
* Several contracts support.

May you give me some thought about these three problems?

These bugs I will fix:
* ```run method``` is not work correctly.

Next, I will split the opcode id to another pr.

peroket:
> > peroket
> 
> Yes, I think your points are right. I think we may need more discussion to solve these problems.
> 
> * The `run` need to be split into two methods, `` compiler`` and  ``run```.
> * The arguments format is `args1,args2,args3`, this format is the same as the remix.
> * Several contracts support.
> 
> May you give me some thought about these three problems?
> 
> These bugs I will fix:
> 
> * `run method` is not work correctly.
> 
> Next, I will split the opcode id to another pr.
> 
> Thank you.

I'm not really good at UI, but I don't think we can just add buttons in the current one. Each windows are already pretty small, and adding buttons is just taking more space. So probably to add all that the UI would need to be revamped, like for example having a different UI for writing code and for running a transaction. That would make the first point more logical as well, and third easier maybe.

For the arguments, do you have a link? That should be available from the UI as well. The types supported, the way to write the data...

Of course I'm talking ideally what I would like to have :-)

jonahcui:
We can use ```ethereumjs-abi``` to convert the method's signature and the arguments as follows:

When we have a contract with a method ```inc``` that needs two arguments which type are both ```uint256```.  So we can use ```abi.simpleEncode("inc(uint256, uint256)", 12, 13)``` to encode this data to ```0xe8a7201f000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000d```.

In ```SolidityMethodSelector``` component, you can select ```inc()``` method. Then fill ```12,13``` into arguments input. The component use ```,``` to split the user's input. The pageshot as follows.
CleanShot 2022-06-18 at 11 20 50
CleanShot 2022-06-18 at 11 21 12


> The ```ethereumjs-abi```  support most of solidity argument types, such as: ```addresss```, ```bool```, ```uintN```, ```intN``` and ```[]``` etc.

I agree with your point that the window is small enough. we can't add more buttons to this page.

How about adding some ```Dialog``` or ```Modal```on this page, then we can let the users select contract, select method, change argument, and change value? 

If we add a new playground, users may feel confused about the design. 

Now, we can follow the step to run a solidity method:
* select ```solidity``` code type.
* Then, run the contract
* If the contract constructor has some arguments, we need to add the argument. And click the ```run``` button again. 
* Next, click the ``` step into``` button or ```continue execution``` button. It will deploy the contract.
* Next, we can choose a contract method, and put the argument or the value
* Next, click ```run method``` button
* Finally, click the ``` step into``` button to debug your method.

> After deploying the contract, the transaction will return the opcodes about the contract methods. The opcodes can be used to call ```loadInstructions``.

I think these are so complex that I have no idea about cutting down some steps. Do you have some ideas?

dorlevi:
Yeah, I guess one way would be to add an "Advanced" option that opens up a modal/dialog box with these additional settings, another option is to have that advanced settings expand over the console and then minimize when you run/click minimize. 

A VERY UGLY illustrative sketch, that way you have more room without another dialogue box/modal, I don't think the console is that useful anyway before you run anything.


Screen Shot 2022-06-18 at 11 32 17 AM


also @thevaizman might have ideas here

jonahcui:
This is a good idea, I will do it

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: Dor Levi ***@***.***>
发送时间: Saturday, June 18, 2022 11:35:50 PM
收件人: comitylabs/evm.codes ***@***.***>
抄送: jonahcui ***@***.***>; Author ***@***.***>
主题: Re: [comitylabs/evm.codes] Feature add solidity method run (PR #148)


Yeah, I guess one way would be to add an "Advanced" option that opens up a modal/dialog box with these additional settings, another option is to have that advanced settings expand over the console and then minimize when you run/click minimize.

A VERY UGLY illustrative sketch, that way you have more room without another dialogue box/modal, I don't think the console is that useful anyway before you run anything.

[Screen Shot 2022-06-18 at 11 32 17 AM]<https://user-images.githubusercontent.com/1457950/174445611-6fb12a79-83c1-4e77-87dd-0b09783ebfbc.png>

also @thevaizman<https://github.com/thevaizman> might have ideas here

―

jonahcui:
@peroket @dorlevi 

Hello, everyone. I had made a solution through our discussion. I agree with @peroket that we don't need more complex buttons in the window. So I make three scenarios, each scenario has one master button in the window.

Fist, Simple Mode:
CleanShot 2022-06-20 at 13 28 32
When I haven't compiler my contract, the window will show the ```run``` button as same as current.

Second, when the user compiles a contract whose constructor has more arguments, I show an error in the console window. As follows:
CleanShot 2022-06-20 at 13 34 45

Then, when I click the ``advance mode `` button, I will hide the ```run``` and show the ``` run method``` bar.
CleanShot 2022-06-20 at 13 36 09
> the button text is Deploy because the contract had not been deployed. I will select the constructor for user automatic.


After deploying the contract, the user can choose another method to run or change the mode to ```Simple Mode```,
CleanShot 2022-06-20 at 13 38 43

peroket:
> @peroket @dorlevi
> 
> Hello, everyone. I had made a solution through our discussion. I agree with @peroket that we don't need more complex buttons in the window. So I make three scenarios, each scenario has one master button in the window.
> 
> Fist, Simple Mode: CleanShot 2022-06-20 at 13 28 32 When I haven't compiler my contract, the window will show the `run` button as same as current.
> 
> Second, when the user compiles a contract whose constructor has more arguments, I show an error in the console window. As follows: CleanShot 2022-06-20 at 13 34 45
> 
> Then, when the user clicks the `advance mode ` button, The run bar will be hidden, and the ` run method` will be shown.
> 
> CleanShot 2022-06-20 at 13 36 09
> 
> > the button text is Deploy because the contract had not been deployed. I will select the constructor for user automatic.
> 
> After deploying the contract, the user can choose another method to run or change the mode to `Simple Mode`, CleanShot 2022-06-20 at 13 38 43
> 
> I will push my code after our discussion. Looking forward to your reply.

This is already a big improvement thanks :-)
However I don't see anything that deals with several contracts deployed in your solution. Do we explicitely accept to have only the last deployed that can be interacted with?
Also, I don't see how it solves the limited space problem, can you explain further? I don't really like the idea of modal windows, because they become cumbersome really quickly.

Feel free to push your code so that we can play with it :-)

jonahcui:
I'm sorry, I didn't solve the several contract problem. I think several contracts are not a common scenario. But I can add a ```contract select``` component in the method select row. If we need to solve it.

I had added the *advanced mode* which is different from the current run mode. The advance mode is hidden in the run bar and creates new three rows in the editor window.
* First row used to be ```select method```, then the user can click ```deploy``` or ```run method```
* The second row used to be setting values like the input in the run bar.
* The other rows used to be setting the argument by the input that shows the argument type 
* The bar only shows three rows once, if the contract has more arguments, the scroll bar will be shown.
*  The bar's height is 52 * 3 (run bar's height is 52)
CleanShot 2022-06-21 at 16 33 33

I pushed my code yesterday.

jonahcui:
Hello, all, I had added a ```contract select```. like this:
CleanShot 2022-06-23 at 15 23 22

If we don't like the scrollbar in *Advance Mode*, I think we can change the arguments input to a single input, like this:
CleanShot 2022-06-23 at 15 27 04

arguments joins with ```,```

Would you like to give me some thoughts about the solution?

peroket:
This is looking better again thanks :-)

My first comment would be that I find it weird that I can't use advanced mode before I click on "run". I know If my contract needs arguments, so I should probably be able to go to advance mode directly.

The logs also give me errors when I deploy in advance mode, even though I put a value and it launched the transaction. And it printed the same error 3 times for 1 click ("The constructor of the contract need arguments...").
And running the transaction to the end again prints the error countless time, even though it did deploy the contract.

Then I ran the "count" method, which ran fine, but after that all the methods are grayed out and I cannot run anything anymore.

I then modified the contract to add two constructor parameters, and changed the name of the contract. I was able to deploy it even though I didn't give any parameter to the contructor, the ui did not update to ask me two parameters, and the new contract did not appear in the contract list. Plus again a lot of error printings.

So I'm sorry, but it looks like it is nowhere near ready to be deployed. Interface wise though, What I could test works well, and I don't think we need anything more for a simple tool like ours anyway, so nothing to do there. Would be nice to have similar features for the other mode (yul...), except there it would simply be select contract address and enter the calldata manually. Also when I am in advanced mode and swith to another language, it does not reset the UI.

I know it's a lot of efforts to do all of that, so don't worry if you can't/dont't want to finish it :-)

jonahcui:
I think the *multiple contract support* and *contract method arguments* make the UI more complex. And I will complete this feature with test all scenarios. But I don't think the feature is not necessary, because we need to show the users how EVM works with arguments when the users are similarly with solidity.

And I think if we add a  page that shows more code examples like [Solidity by example](https://solidity-by-example.org/) or other code example pages. The *evm.codes* might be the best app for evm beginners.

I will fix all those bugs and think of a better solution to cut down the confusion for users.
> deploy contract need to click the ```deploy``` button and then click the ```continue execution``` button, because some users may like to see how does evm deploy a contract. Maybe we need a better solution to do it.  But if I just want to run a method of the contract, it's too complex. Any thoughts on it?

I will try it, but if you have other ideas that need to be implemented, I'll be there. I can do it.

peroket:
For the mutliple contract agreed, I'm fine explicitely supporting only the last deployed contract. I was not asking to support several, just to make a choice that's what we do.

jonahcui:
OK, I will fix all the bugs and change some logic to make it more simple.

jonahcui:
I'm sorry that I hadn't found an excellent way to make users deploy contracts simpler. So I won't close this pr, if some day we find a perfect way to solve the problem, we can try it again.

jonahcui:
If you want to run a contract method, you need to deploy the contract and click the ```continue execute```, then the contract will be deployed successfully. Next, you can run the method that you need.

The process is so confusing. But I didn't know how to make it simpler.

jonahcui:
Relax, I have fixed some bugs, and I will fix the other bugs tomorrow. I will push the branch after all the bugs are fixed. Then we can test it together.
 
But I want to solve the problem of how to deploy a contract.

jonahcui:
@peroket @dorlevi 
CleanShot 2022-07-01 at 15 59 41


I found a method to make users don't feel confusing. I add a compile option in method selection. And if you change to the advance mode when the user has not compiled the contract before, I will autoselect compile option for the user.

The Simple Mode:
* If the user clicks run, I will compile the contract first, then deploy it.
* When the user contract has arguments, the console will show revert after the user debug the transaction.

From The Simple Mode To The Advanced Mode:
* If the user has not compiled the contract, I will autoselect the compile option for the user.
* If the user compiles the contract before, I will autoselect the constructor for the user.
* If the user has deployed the contract before. I will autoselect the constructor for the user. We can talk about whether to help users choose the first method.

The Advanced Mode:
* The users can choose the compile option to compile the code.
* The users can choose the constructor option to deploy the code.
* The users can choose a method to run. But users must deploy the code first.

jonahcui:
@peroket @dorlevi 
CleanShot 2022-07-01 at 15 59 41


I found a method to make users don't feel confusing. I add a compile option in method selection. And if you change to the advance mode when the user has not compiled the contract before, I will autoselect compile option for the user.

The Simple Mode:
* If the user clicks run, I will compile the contract first, then deploy it.
* When the user contract has arguments, the console will show revert after the user debug the transaction.

From The Simple Mode To The Advanced Mode:
* If the user has not compiled the contract, I will autoselect the compile option for the user.
* If the user compiles the contract before, I will autoselect the constructor for the user.
* If the user has deployed the contract before. I will autoselect the constructor for the user. We can talk about whether to help users choose the first method.

The Advanced Mode:
* The users can choose the compile option to compile the code.
* The users can choose the constructor option to deploy the code.
* The users can choose a method to run. But users must deploy the code first.

jonahcui:
@peroket @dorlevi 
I found a method to make users don't feel confusing. I add a compile option in method selection. And if you change to the advance mode when the user has not compiled the contract before, I will autoselect compile option for the user.
CleanShot 2022-07-01 at 16 00 52


The Simple Mode:
* If the user clicks run, I will compile the contract first, then deploy it.
* When the user contract has arguments, the console will show revert after the user debug the transaction.

From The Simple Mode To The Advanced Mode:
* If the user has not compiled the contract, I will autoselect the compile option for the user.
* If the user compiles the contract before, I will autoselect the constructor for the user.
* If the user has deployed the contract before. I will autoselect the constructor for the user. We can talk about whether to help users choose the first method.

The Advanced Mode:
* The users can choose the compile option to compile the code.
* The users can choose the constructor option to deploy the code.
* The users can choose a method to run. But users must deploy the code first.

jonahcui:
Thanks, I have finished these bugs:
* Remove the default constructor in the example of solidity.
* Can not click ```run``` when returning from advance mode.
* The ``` run method with``` tips has been changed to three tips:
  * if the method has arguments and the user inputs arguments, then the console will show ```run method with arguments.```
  * if the method has no arguments and the user inputs arguments, then the console will show ```Ignore arguments```
  * if the method has no arguments and the user doesn't input arguments, then print ```run method.```
 * Add width for inputs. And add a title for the arguments line.
 * Change text-transform to 'none' for the method name in select.
 * Add "Share permalink" button after method selected.

Some question:
* Whether we need to scroll to the bottom for the console after pushing a log?

peroket:
Scrolling to the bottom would be very convenient yes, if you can do that I would be grateful :-)

peroket:
Thanks @jonahcui as a user looks great :-)
Will get a technical review soon as well.

components/Editor/index.tsx
```diff
@@ -119,9 +127,46 @@ const Editor = ({ readOnly = false }: Props) => {
...
+  const deployByteCode = useCallback(
+    (bc, args = '', callValue) => {
+      try {
+        if (!callValue) {
+          callValue = getCallValue()
+        }
+        return transactionData(bc + args, callValue).then((tx) => {
+          loadInstructions(bc)
+          setIsCompiling(false)
+          return startTransaction(tx).then((result) => {
```
reichert621, line 139: probably best not to nest this promise in here like this -- we could either chain it and then explicitly handle a `.catch`, or we could just convert this whole outer function to be `async`, e.g.

```typescript
const deployByteCode = useCallback(
  async (bc: string, args: string, callValue: BN | undefined) => {
    try {
      if (!callValue) {
        callValue = getCallValue()
      }
      const tx = await transactionData(bc + args, callValue)
      loadInstructions(bc)
      setIsCompiling(false)
      const result = await startTransaction(tx)

      if (
        codeType === CodeType.Solidity &&
        !result.error &&
        result.returnValue
      ) {
        setMethodByteCode(bufferToHex(result.returnValue))
      }
      return result
    } catch (error) {
      log((error as Error).message, 'error')
      setIsCompiling(false)
    }
  },
  [
    transactionData,
    getCallValue,
    loadInstructions,
    startTransaction,
    codeType,
    log,
  ],
)
```
this way we `catch` any errors from all promises in the `try` block :)

components/Editor/SolidityAdvanceModeTab.tsx
```diff
@@ -0,0 +1,410 @@
...
+    if (isEmpty(data)) {
+      return
+    }
+
+    transactionData(
+      data,
+      getCallValue(),
+      Address.fromString(deployedContractAddress),
+    ).then((txData) => {
+      startTransaction(txData).then((result) => {
```
reichert621, line 208: nit -- might be better to avoid nesting these promises, and add some error handling :)

reichert621 (commented):
just a couple comments!

one UX suggestion -- when running the "deploy"/"compile"/"run" button, it would be nice to add some visual "pending" feedback (e.g. a spinner on the button) so i know that something is happening. 

while testing, i was having trouble deploying because it was just hanging, but from a UI perspective it seemed like nothing was happening (e.g. i clicked "deploy" and received no visual feedback)

components/Editor/index.tsx
jonahcui: Ok, I will use ```async``` and ```await``` to change the ```promise```

jonahcui:
> 

I try to add a ``` running``` state to implement the UI suggestion about button feedback. Buut the run time is so short that the user can't see the button style changed.

Finally, I added a ```active:opacity-50``` to the button's class name.

components/Editor/SolidityAdvanceModeTab.tsx
jonahcui: Thanks, :D. I have fixed it.

peroket (approved):

jonahcui:
I have fixed some bugs, and I will fix the other bugs tomorrow. I will push the branch after all the bugs are fixed. Then we can test it together.


But I want to solve the problem of how to deploy a contract.

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: Dor Levi ***@***.***>
发送时间: Thursday, June 30, 2022 6:58:29 PM
收件人: comitylabs/evm.codes ***@***.***>
抄送: JonahCui ***@***.***>; Mention ***@***.***>
主题: Re: [comitylabs/evm.codes] Feature add solidity method run (PR #148)


@jonahcui<https://github.com/jonahcui> : We appreciate the work here and understand you want to merge it (or do you mean abandoning and closing it?) We can punt on the ease of use of deploying a contract, but can you take care of some of the bugs mentioned earlier by @peroket<https://github.com/peroket> ? at this stage, I don't believe it's not ready to review/merge; it's a slippery slope to merge stuff that is not entirely up to the standard of experience we would want for the users of evm.codes _and as the branch is automatically deployed by Vercel so if someone wants to use this functionality before it's merged they can do that easily so we rather not merge it until some of the larger bugs are addressed.

If you can't address the bugs, that is fine, you've done a lot already -- we are soon onboarding additional FE engineers to our team (evm.codes is not our main product) and have them take it from here when they start in a couple of weeks.

―

drinkius:
Got an argument list as follows:
```
solution(uint256[2][3],uint256[2][3])
```
Can't figure out the correct way to supply the arguments. What works with Remix - `[[1, 1], [1, 1], [1, 1]], [[2, 6], [0, 10], [3, 3]]` - doesn''t work with EVM.codes Playground

PR #149, opened 2022-06-18
[chore] add opcode id show.

jonahcui:
We should add the opcode id in the Instruction row, so we can trace the next opcode id when the current opcode is```JUMP``` or ```JUMPI```.

CleanShot 2022-06-18 at 10 59 23


malexw (approved):
Seems like a good idea to me - thanks!

Issue #145, opened 2022-06-14
[playground] Switching from bytecode to mnemonic (and viceversa and other combination) should "translate" to the destination language


StErMi:
Hi there, 

As a dev/user, I would expect that if I've inputted a bytecode inside the playground, run it, and then I switch to the mnemonic version it would show me the mnemonic version of the bytecode.

Currently, it does not that, and it forces me to go to an external tool, translate it and then paste it again. 
It would be really nice if this feature is enabled by default for all the various language  (yul, solidity, bytecode, mnemonic)

peroket:
That's a very good idea yes, thank you.
I don't think we can support conversions to yul or solidity, but to opcodes and mnemonic definitely.

jonahcui:
@malexw Had you started to do this feature? if you don't have time, I can do this.

malexw:
@jonahcui Nope, I haven't had a look at it. It's all yours!

peroket:
> @peroket @dorlevi I had finished this in the commit [7963495](https://github.com/jonahcui/evm.codes/commit/79634959a4b2285afc8b4e095a122589b68c6713)
> 
> But I found a problem:
> 
> * I compile the `yul` contract and run it.
> * Then, I change the code type to bytecode and do nothing.
> * Then, I change the code type to the mnemonic. I will change the code in the editor to the opcodes of the contract that I compile the `yul` contract.
>   Same issues when switching to bytecode from mnemonic.
> 
> Whether we need to fix this problem? If we don't need to fix this problem. I will pull the request.

I'm not sure I understand the problem? If when you switch it changes opcode to mnemonics and back, sounds like what we want. Maybe better though, having a button "translate", that will show the possible options. So for solidity and yul, there would be opcodes and mnemonics, and for opcode and mnemomics, jst the other one. That way everything is explicit and it's not weird when I switch from opcode to yul for example.

What do you think? Also feel free to open a PR so that we can play with it :-)

jonahcui:
Oh
It just means I will translate yul to mnemonics or bytecode. The code will show Yul's bytecode. When you first switch to yul, then switch to mnemonics, and then switch to bytecode.

devtooligan:
thanks for workimg on this @jonahcui

[...]

PR #152, opened 2022-07-01
[feat] Switching from bytecode to mnemonic (#145)

jonahcui:
Hi, all. This pr will add support for translate between bytecode and mnemomics.


dorlevi:
thanks again for your contributions!

Just played around with this branch and I'm getting an [Error] Invalid EVM version requested when running/deploying a contract, will look take another look later to find the issue

jonahcui:
ok, I'll fix it soon.

components/Editor/index.tsx
```diff
@@ -294,6 +298,26 @@ const Editor = ({ readOnly = false }: Props) => {
...
+      value &&
+      lastCodeType === CodeType.Bytecode &&
+      value === CodeType.Mnemonic &&
+      instructions?.length > 0
+    ) {
+      const code = getBytecodeLinesFromInstructions(instructions)
+      setCode(code)
+    }
+
+    if (
```
assimovt, line 311: should this be else if?

components/Editor/index.tsx
```diff
@@ -294,6 +298,26 @@ const Editor = ({ readOnly = false }: Props) => {
       setCode(examples[value as CodeType][0])
     }
 
+    if (
```
assimovt, line 301: should this be else if

components/Editor/index.tsx
```diff
@@ -284,6 +287,7 @@ const Editor = ({ readOnly = false }: Props) => {
 
   const handleCodeTypeChange = (option: OnChangeValue<any, any>) => {
     const { value } = option
+    const lastCodeType = codeType
```
assimovt, line 290: do you really need to redefine codeType?

components/Editor/index.tsx
jonahcui: I've fixed it.

components/Editor/index.tsx
jonahcui: I've fixed it.

components/Editor/index.tsx
jonahcui: I remove the ```codeType``` and the statement that checks whether ``` lastCodeType``` is equal to bytecode or mnemonic in the if statement.

peroket:
Hello, thanks for your contribution :-)
I think it's good, but from a user point of view I think it might be confusing, because first if you don't know it exists there's nothing telling you it will do that, and second because there is nothing to say if it actually worked or not.

So to integrate properly this feature, I think idealy it would be only on explicit request from the user, with possibly an error message in the console if it's not possible for some reason (like unknwon opcode). For the UI integration, I'm thinking next to the current drop down menu to switch language, adding another drop down menu with a "Convert to: X". Currently the option would be mnemonic to opcodes or the reverse, but that way we can imagine adding more, like solidity to any other and yul to opcode or mnemonic as well.

I know it's a lot to ask, so no worries if you can't do it :-)

dorlevi:
thanks for making the fixes @jonahcui. I agree with @peroket  that to make this integrated and complete giving some indication if the feature worked in the console would be good. however, I don't believe we need an explicit dropdown just for this feature, in fact, I wouldn't want one even if you worked on it, so @jonahcui , don't worry about that.

@peroket, this feature is not going to be wildly used (assumption) or the main use case, so taking space in the UI and creating some clutter/more confusion for the main usecases for a secondary one is not something I'd like to do , also evm.codes is a pro-user tool, it's fine if there are some features that are not as discoverable and users can just find them by playing around or reading release notes/seeing on twitter, so all and all I prefer not to make it's own dropdown or anything like that, I can think about solutions that might both achieve discoverability while not adding more UI elements that won't be used often, such as a short console message "you can now switch between XXX and aYYY" or as a small banner but I wouldn't even bother on these. 

So to summarize, let's properly integrate this into the console and that's it.

P.S I am not too concerned that switching will not require an "explicit" action but will just work when switching the language selector, but part of how I would want to implement the console messages is that if it DOES fail, we will revert to showing the default mnemonic example, I would want to avoid situations where people switch to the mnemonic display and don't see anything because they had some malformed bytecode if that make sense.

jonahcui:
@peroket @dorlevi OK, I see all the information. the key points are:
1. I agree that the function is not common, maybe it's just a small improvement
2. Add another dropdown is a good solution, but it can make users feel confused about why does there have two dropdowns. So I prefer the idea that adding a console message to adding another dropdown.

dorlevi (approved):

dorlevi (approved):

PR #157, opened 2022-07-24
chore: fix the playground example for CREATE

OlivierBBB:
The bytecode `0x63FFFFFFFF60005260046000F3` that is deployed in the 3rd CREATE in the CREATE playground is strange in the sense that it PUSHes `0xFFFFFF` on stack, MSTOREs it and returns 4 bytes (RETURN instruction with offset 0 and size 4). However this returns `0x000000`. This isn't a mistake per se, it seems like a typo where the intent was to return the 4 bytes previously MSTOREd.

This pull request changes the bytecode to `63FFFFFFFF6000526004601CF3` so that the returned bytecode coincides with the 4 bytes previouslypushed on stack.

### DIFF
--- a/docs/opcodes/F0.mdx
+++ b/docs/opcodes/F0.mdx
@@ -35,7 +35,7 @@ Note that these failures only affect the return value and do not cause the calli
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


peroket (approved):
Thanks :-)

PR #163, opened 2022-08-10
SEO 3

sekrystal:
- Reformatted “about” page to effective metatag syntax
- Lengthened “home” and “playground” page meta-descriptions for visibility

### DIFF
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -21,9 +21,9 @@ const SectionWrapper: React.FC<SectionWrapperProps> = ({
     <>
       <Head>
         <meta property="og:type" content="website" />
-        <meta property="og:title" content="EVM Codes - About the EVM" />
+        <meta property="title" content="EVM Codes - About the EVM" />
         <meta
-          property="og:description"
+          name="description"
           content="Demystifying what makes the Ethereum Virtual Machine work."
         />
       </Head>
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -1,3 +1,5 @@
+import React from 'react'
+
 import type { NextPage } from 'next'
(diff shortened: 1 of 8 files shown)


pages/about.tsx
```diff
@@ -19,14 +21,17 @@ const SectionWrapper: React.FC<SectionWrapperProps> = ({
...
-          name="description"
-          content="Demystifying what makes the Ethereum Virtual Machine work."
-        />
-      </Head>
+      <html lang="en"></html>
+      <React.Fragment>
+        <Head>
+          <title> EVM Codes - About the EVM </title>
+          <h1>
+            Understanding the Ethereum Virtual Machine (or EVM) is understanding
```
dorlevi, line 29: I don't love this language, it's too much like a fortune cookie a proposal "Understand the relationship between opcode instructions, gas costs, storage, and the execution environment to better understand the EVM" or something

pages/playground.tsx
```diff
@@ -8,9 +8,10 @@ import { Container } from 'components/ui'
 const PlaygroundPage = () => {
   return (
     <>
+      <html lang="en"></html>
       <Head>
         <meta property="og:type" content="website" />
-        <meta property="title" content="EVM Codes - Playground" />
+        <title>EVM Codes - Playground</title>
```
dorlevi, line 14: You don't want to add something here EVM Codes - Explore your contracts in our Playground" or something?

sekrystal: That'd be too long – the longest tagline we'd want is for the homepage. This particular text is what you'd see in a tab bar, the shorter, the better.

dorlevi (approved):
LGTM

PR #165, opened 2022-08-18
Fix byte range of `ecadd` precompile

mriise:
#141 fixed a typo of the mislabeled point y2, but missed the byte range


peroket (approved):
Thank you :-)

PR #166, opened 2022-08-26
Fix typo in byte range of precompiled contracts

minaminao:
Changed from `[0; length[` to `[0; length]`

### DIFF
--- a/docs/precompiled/0x03.mdx
+++ b/docs/precompiled/0x03.mdx
@@ -6,7 +6,7 @@ fork: Frontier
 
 | Byte range | Name | Description |
 |-----------:|-----:|------------:|
-| `[0; length[` | data | Data to hash with RIPEMD-160 |
+| `[0; length]` | data | Data to hash with RIPEMD-160 |
 
 ## Output
 
--- a/docs/precompiled/0x04.mdx
+++ b/docs/precompiled/0x04.mdx
@@ -10,13 +10,13 @@ The identity function is typically used to copy a chunk of memory.
 
 | Byte range | Name | Description |
 |-----------:|-----:|------------:|
-| `[0; length[` | data | Data to return |
+| `[0; length]` | data | Data to return |
 
 ## Output
 
 | Byte range | Name | Description |
 |-----------:|-----:|------------:|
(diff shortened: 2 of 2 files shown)


peroket (commented):
Sorry but this is not a typo, this is to specify that the range is up to but not including the length
Thank you for proposing it though, all contributions are welcome :-)

minaminao:
Ah, I see! Sorry for the unnecessary suggestion. I will close this PR. 
By the way, I have never seen this kind of notation before, is it commonly used? 
I would appreciate any resources written about this notation for future reference.

peroket:
> Ah, I see! Sorry for the unnecessary suggestion. I will close this PR. By the way, I have never seen this kind of notation before, is it commonly used? I would appreciate any resources written about this notation for future reference.

I actaully don't knoy if it is commonly used, it's a math thing, I assumed it was.

minaminao:
Thank you. I only knew how to indicate that one of the endpoints is to be excluded from the set as `[x,y)` in math, but I just learned that it is also expressed as `[x,y[` by the following pages fyi:

- https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints
- https://cdn.standards.iteh.ai/samples/3653/0148c25005744dc1ad52152b3d6f469b/ISO-31-11-1992.pdf

image

peroket:
> Thank you. I only knew how to indicate that one of the endpoints is to be excluded from the set as `[x,y)` in math, but I just learned that it is also expressed as `[x,y[` by the following pages fyi:
> 
> * https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints
> * https://cdn.standards.iteh.ai/samples/3653/0148c25005744dc1ad52152b3d6f469b/ISO-31-11-1992.pdf
> 
> image

If the parenthesis notation is more common, then I'm happy to approve that change :-)
Thanks for looking it up

minaminao:
> If the parenthesis notation is more common, then I'm happy to approve that change :-)

I'm Japanese and `[x,y)` is common in Japan, but `[x,y[` may be more common globally, so I don't know which is better :)

PR #171, opened 2022-09-13
Bump Solidity version to 0.8.17

ivigamberdiev:


peroket (changes requested):
It seems however that more changes are needed to support this version, it's not working for me.

ivigamberdiev:
Yes, you're right. For some reason, there are no issues with 0.8.16 but something wrong with 0.8.17

thevaizman:
> Yes, you're right. For some reason, there are no issues with 0.8.16 but something wrong with 0.8.17

@ivigamberdiev Taking a look at the error I get using your build it looks like a simple error in `solcWorker.js`:
image

I will close this PR, fix the error and open a new PR with 0.8.17 as the version of `soljson`.

PR #173, opened 2022-09-21
Add fork to query params

thevaizman:
See this [issue](https://github.com/comitylabs/evm.codes/issues/84)

Added the option to provide a `fork` query parameter to the following pages:

- Precompiles
- Playground
- Reference Table

Also added the `fork` query parameter to the permalink copy feature of Playground.

### DIFF
--- a/components/ChainSelector.tsx
+++ b/components/ChainSelector.tsx
@@ -1,7 +1,7 @@
 import { useContext, useEffect, useMemo, useState, useCallback } from 'react'
 
-import { useRouter } from 'next/router'
 import { useRegisterActions, Action } from 'kbar'
+import { useRouter } from 'next/router'
 import Select, { OnChangeValue, components } from 'react-select'
 
 import { EthereumContext } from 'context/ethereumContext'
@@ -51,6 +51,7 @@ const ChainSelector = () => {
       router.query.fork = option.value
       router.push(router)
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     [onForkChange, setSetting],
   )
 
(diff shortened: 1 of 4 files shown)


dorlevi:
Currently the PR enables you to use query params but it doesn't actually use it when you click the "link" button on an opcode or just when you browse the latest fork, should that change?

thevaizman:
> Currently the PR enables you to use query params but it doesn't actually use it when you click the "link" button on an opcode or just when you browse the latest fork, should that change?

I wasn't sure if this is wanted behavior or not, but I guess it might be useful for sharing opcodes/precompiles since they don’t have a “copy permalink” button like the playground.
Will add that.

dorlevi:
Yeah, I could see reasons to go either way on it.  so totally fine keeping it as is/simpler

thevaizman:
> Yeah, I could see reasons to go either way on it. so totally fine keeping it as is/simpler

Already committed the change :)
I see the value in adding the fork to the URL to make sharing URLs between users more “transparent”.

components/ChainSelector.tsx
```diff
@@ -45,7 +47,11 @@ const ChainSelector = () => {
       setForkValue(option)
       onForkChange(option.value)
       setSetting(Setting.VmFork, option.value)
+
+      router.query.fork = option.value
+      router.push(router)
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     [onForkChange, setSetting],
```
dorlevi, line 55: ?

thevaizman: This line simply tells the linter to disable the `react-hooks/exhaustive-deps` rule for the line following it.
It is pretty common to disable this warning as it sometimes warns you about a missing dependency but that dependency is unnecessary (this time it was warning about the `router` which cannot even be declared inside this `useCallback` block...

dorlevi (approved):
to my untrained eye this seems fine

Issue #122, opened 2022-03-11
Feature request: support URL query strings for opcode searches


maurelian:
First, this site is amazing, thank you for continuing to improve it!

**Request:** I would _love_ to be able automate a quick opcode (or precompile) search using a format like `https://www.evm.codes/search?fork=arrow-glacier&opcode={query}`. Where the fork defaults to the current `live` value.

I'd use this site on a daily basis if it were that easy. :)

dorlevi:
Yep, I see how this would be handy to quickly search without using the site, we are pretty heads down on our core product but we will keep this in mind, seems like a good starter project to onboard a frontend/full-stack while letting them read more about the EVM :) 

If there is a front/end-full-stack engineer who sees it and wants to pick this up we will compensate them, in general, if you know someone who would want to help us improve this and work  on some other opensource utilities free we have in mind on a part-time or even fulltime basis we would love to talk to them

Saw-mon-and-Natalie:
I would love to help with this feature. 

@maurelian, do you want that endpoint to work kind of as an API? Like return data in `JSON` format. Also what date you would it to return? (opcode, name, gas, stack slots consumed, stack slots produced, description, ...)

peroket:
Thanks @Saw-mon-and-Natalie 
I guess the first step would be to add the fork in the link to opcodes, because I realise now that it is missing. That should not be too hard, if you want to start with that?

For example when I click on the first link to opcode 00, I get https://www.evm.codes/#00, would be even better with the fork.

Saw-mon-and-Natalie:
Should the link be like ` https://www.evm.codes/#00?fork=arrow-glacier` even though the `fork` query is not being used?

peroket:
The idea would be to add also setting the fork according to the parameter (or leave the current behaviour if there is none).

iFrostizz:
any progress on this feature ?

dorlevi:
we have not worked/prioritized it, but if someone submits a PR we will review and merge it.

thevaizman:
@maurelian I merged a PR that contains the changes described in your original post :)
Closing this issue for now.

maurelian:
Amazing, TY!

PR #175, opened 2022-09-25
Added `name` query parameter handling

thevaizman:
Added a `name` query parameter handling code to the `Filters.tsx` component. This handler will set a filter on the `Name` column of a table (either opcodes or precompiles) with the value of the query parameter.
In example:

- `evm.codes/precompiled?name=ec` will yield a search on the precompiles table for precompiles containing the sequence `ec` (ecrecover, ecpairing).
- `evm.codes/?name=self` will yield a search on the opcodes table for opcodes containing the sequence `self` (SELFDESTRUCT, SELFBALANCE).

Note: I did not set an `opcode` query parameter with the opcode's hex form (e.g `0x60`) since we already have the anchored views for that.

This PR should close this issue #122

### DIFF
--- a/components/Reference/Filters.tsx
+++ b/components/Reference/Filters.tsx
@@ -51,7 +51,7 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
 
     if ('name' in query) {
       // Change the filter type to Name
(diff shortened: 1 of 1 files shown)


dorlevi (approved):
LGTM

PR #176, opened 2022-09-27
updated all reference of /about to the correct internal section of the about page

StErMi:
Inside this PR I have updated all the `/about` reference to the specific section inside the About page. 

For example 

```diff
-[calldata](/about)
+[calldata](/about#calldata)
```

## Important note

Don't ask me why, but my prettier/dunno what have sometimes auto-formatted some markdown section :(

### DIFF
--- a/docs/opcodes/00.mdx
+++ b/docs/opcodes/00.mdx
@@ -5,6 +5,6 @@ group: Stop and Arithmetic Operations
 
 ## Notes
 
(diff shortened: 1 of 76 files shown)


thevaizman (approved):
Looks good. See my [comment](https://github.com/comitylabs/evm.codes/pull/176#issuecomment-1259537534).

PR #177, opened 2022-09-29
Fix gas reference for SSTORE

ernestognw:
## Description

After #176, the `SSTORE` reference just broke due to a lack of indentation.
This PR fixes that

### Before
Captura de Pantalla 2022-09-29 a la(s) 12 16 51

### After
Captura de Pantalla 2022-09-29 a la(s) 12 16 57

### DIFF
--- a/docs/opcodes/55/berlin.mdx
+++ b/docs/opcodes/55/berlin.mdx
@@ -6,20 +6,21 @@
 - **current_value**: current value of the storage slot.
 - **original_value**: value of the storage slot before the current transaction.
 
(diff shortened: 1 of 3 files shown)


thevaizman:
Hey @ernestognw, thanks for alerting us! It appears as though some auto-formatting that was introduced in (#176) broke the indentation of several files (not just `SSTORE`).
I opened a [PR](https://github.com/comitylabs/evm.codes/pull/178) that simply reverts the broken commit so I will close this one and we will merge the reverting PR instead.

Again, thanks for alerting! 🙏

PR #178, opened 2022-09-29
Revert broken commit

thevaizman:
Revert "updated all reference of /about to the internal about section (#176)"

This reverts commit 49fecec35cf816cbf5507a89ad8d79b3f1e9d849.

Reverting commit due to indentation problems. Will open a new PR that introduces only the "safe" changes of this commit.


dorlevi (approved):
+1

PR #180, opened 2022-10-12
Replace difficulty with prevrandao

thevaizman:
This PR contains quite a lot of code that had to be refactored in order to use the latest `ethereumjs` library (which supports the merge).
In addition a hacky prebuild script was added to the build process. Details for why we did that are [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2295).

Please review this thoroughly :)


util/compiler.ts
```diff
@@ -16,13 +16,7 @@ export const getTargetEvmVersion = (forkName: string | undefined) => {
   if (forkName === 'dao') {
     return 'homestead'
   }
-  if (forkName === 'muirGlacier') {
```
peroket, line 19: Why did muiGlacier became london?

components/Editor/index.tsx
```diff
@@ -117,17 +117,18 @@ const Editor = ({ readOnly = false }: Props) => {
   )
 
   const getCallValue = useCallback(() => {
-    const _callValue = new BN(callValue)
+    const _callValue = BigInt(callValue)
+    let cv = BigInt.asIntN(32, BigInt(0))
 
     if (unit === ValueUnit.Gwei) {
-      _callValue.imul(new BN('1000000000'))
+      cv = BigInt.asIntN(32, _callValue * BigInt('1000000000'))
```
darkdarkdragon, line 124: What's the reason for this manipulation?
Why do you clamp result to 32 bits?

context/ethereumContext.tsx
```diff
@@ -41,15 +43,15 @@ const privateKey = Buffer.from(
 )
 const accountBalance = 18 // 1eth
 const accountAddress = Address.fromPrivateKey(privateKey)
-const contractAddress = Address.generate(accountAddress, new BN(1))
-const gasLimit = new BN(0xffffffffffff)
+const contractAddress = Address.generate(accountAddress, BigInt(1))
```
darkdarkdragon, line 46:
```suggestion
const contractAddress = Address.generate(accountAddress, 1n)
```

components/Editor/index.tsx
```diff
@@ -117,17 +117,18 @@ const Editor = ({ readOnly = false }: Props) => {
   )
 
   const getCallValue = useCallback(() => {
-    const _callValue = new BN(callValue)
+    const _callValue = BigInt(callValue)
+    let cv = BigInt.asIntN(32, BigInt(0))
```
darkdarkdragon, line 121:
```suggestion
    let cv = 0n
```

components/Editor/index.tsx
darkdarkdragon, line 124:
```suggestion
      cv = _callValue * BigInt('1000000000')
```

components/Editor/index.tsx
```diff
@@ -117,17 +117,18 @@ const Editor = ({ readOnly = false }: Props) => {
...
 
     if (unit === ValueUnit.Gwei) {
-      _callValue.imul(new BN('1000000000'))
+      cv = BigInt.asIntN(32, _callValue * BigInt('1000000000'))
     } else if (unit === ValueUnit.Finney) {
-      _callValue.imul(new BN('1000000000000000'))
+      cv = BigInt.asIntN(32, _callValue * BigInt('1000000000000000'))
```
darkdarkdragon, line 126:
```suggestion
      cv = _callValue * BigInt('1000000000000000')
```

components/Editor/index.tsx
```diff
@@ -117,17 +117,18 @@ const Editor = ({ readOnly = false }: Props) => {
...
     } else if (unit === ValueUnit.Ether) {
-      _callValue.imul(new BN('1000000000000000000'))
+      cv = BigInt.asIntN(32, _callValue * BigInt('1000000000000000000'))
```
darkdarkdragon, line 128:
```suggestion
      cv = _callValue * BigInt('1000000000000000000')
```

components/Editor/index.tsx
```diff
@@ -117,17 +117,18 @@ const Editor = ({ readOnly = false }: Props) => {
...
     }
 
-    return _callValue
+    return BigInt(cv)
```
darkdarkdragon, line 131:
```suggestion
    return cv
```

context/ethereumContext.tsx
```diff
@@ -41,15 +43,15 @@ const privateKey = Buffer.from(
...
+const gasLimit = BigInt.asUintN(32, BigInt(0xffffffffffff))
```
darkdarkdragon, line 47:
```suggestion
const gasLimit = 0xffffffffffffn
```

context/ethereumContext.tsx
```diff
@@ -478,26 +514,43 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
...
-    vm.stateManager.clearContractStorage = _clearContractStorage
     storageMemory.clear()
   }
 
   const _setupAccount = () => {
     // Add a fake account
     const accountData = {
       nonce: 2,
-      balance: new BN(10).pow(new BN(accountBalance)),
+      balance: BigInt(10 ** accountBalance),
```
darkdarkdragon, line 553:
```suggestion
      balance: 10n ** BigInt(accountBalance),
```

context/ethereumContext.tsx
```diff
@@ -553,9 +606,9 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
     return Block.fromBlockData(
       {
         header: {
-          baseFeePerGas: new BN(10),
-          gasLimit: new BN(0xffffffffffff),
-          gasUsed: new BN(60),
+          baseFeePerGas: 10,
+          gasLimit: BigInt.asUintN(32, BigInt(0xffffffffffff)),
```
darkdarkdragon, line 610:
```suggestion
          gasLimit,
```

util/gas.ts
```diff
@@ -626,8 +626,8 @@ export const parseGasPrices = (common: Common, contents: string) => {
     }
 
     if (namespace === 'gasPrices') {
-      const gasPrice = common.param('gasPrices', key)
-      return gasPrice
+      const gasPrice = Number(common.param('gasPrices', key))
```
darkdarkdragon, line 629:
```suggestion
      const gasPrice = common.param('gasPrices', key)
```

components/Editor/index.tsx
thevaizman: The reason for removing uses of `BN` in favor of `BigInt` in that file is because of the migration to the new version of `@ethereumjs` which dropped support for `BN`.
As for the clamping, totally by mistake, should have been 256 actually (I probably misread bits for bytes 🤦‍♂️ ) will fix and commit, thanks!

context/ethereumContext.tsx
thevaizman: I'm getting a runtime error when committing this change:
`TypeError: Cannot convert a BigInt value to a number`

darkdarkdragon: I see. It compiles it into call `Math.pow(10n, BigInt(accountBalance))`, and `Math.pow` expects `number`.

docs/opcodes/44_merge.mdx
```diff
@@ -0,0 +1,22 @@
+---
+fork: Merge
```
ethosdev, line 2: I think this filename and references to the Merge should be called Paris.

https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md

thevaizman: I totally agree with you, and initially this was the name I was using.
The caveat with using that name, was that the core library we use in our app - `@ethereumjs` uses the name `merge` to refer to the PoS transition. (See [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/a7bb5ba8284894044bbb19f91f3804e901e6d2a0/packages/common/src/hardforks))
I eventually decided to go with their nomenclature in order to avoid mixed references and naming in our app, which IMO is worse than having an inaccurate but consistent references.

Opened [an issue](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2360) on their GH, I invite you to +1 it, and when they complete their renaming, we can too :)

darkdarkdragon (dismissed):
Can't test everything manually, but changes themselves looks good 😄

docs/opcodes/44_merge.mdx
ethosdev: Got it. I +1 the issue you opened. Thanks for explaining.

util/gas.ts
```diff
@@ -641,27 +641,32 @@ export const parseGasPrices = (common: Common, contents: string) => {
...
   )
 
   // filter all forks with block number below or equal to the selected,
   // sort in descending order and pick the first found
+  // Hardcode "merge" block transition block's height as it's determined by TTD and not block height
   let foundFork: string = forkNames
     .filter(
       (forkName) =>
-        knownForksWithBlocks[forkName] <= (selectedFork?.block || 0),
+        knownForksWithBlocks[forkName] <= (selectedFork?.block || selectedFork?.ttd ? 15537394 : 0 || 0),
```
assimovt, line 669: nit: The last || 0 is not needed as it will always default to 0 with this change

prebuild.sh
```diff
@@ -0,0 +1,8 @@
+# This is a HACKY script to fix build errors that occur when using ethereumjs v6+
+# Please refer to this issue if you don't understand what this is all about:
+# https://github.com/ethereumjs/ethereumjs-monorepo/issues/2295#issue-1377817460
+# Hopefully this will be removed when the issue is properly resolved.
```
assimovt, line 4: nit: Is there already an upcoming release branch mentioned [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2295#issuecomment-1278336434) we could reference here to keep an eye on?

prebuild.sh
```diff
@@ -0,0 +1,8 @@
...
+
+sed -i "s/import AsyncEventEmitter = require('async-eventemitter')/import AsyncEventEmitter from 'async-eventemitter'/g" ./node_modules/@ethereumjs/evm/src/evm.ts
+sed -i 's/export .*/export {   getActivePrecompiles, precompiles,   ripemdPrecompileAddress }\nexport type {   AddPrecompile,   CustomPrecompile,   DeletePrecompile, PrecompileFunc,   PrecompileInput }/g' ./node_modules/@ethereumjs/evm/src/precompiles/index.ts
+sed -i '205,$d' ./node_modules/@ethereumjs/evm/src/precompiles/index.ts
```
assimovt, line 8: I think it'd be better to create a `patch` file in the source tree we can apply here instead of relying on sed. Might be cleaner and who know when the new release with the fix will come out.

package.json
```diff
@@ -41,10 +46,15 @@
     "react-select": "^5.1.0",
     "react-simple-code-editor": "^0.11.2",
     "react-table": "^7.7.0",
-    "react-tooltip": "^4.2.21"
+    "react-tooltip": "^4.2.21",
+    "react-usestateref": "^1.0.8"
```
assimovt, line 50: why do we need this one, dont think its coming from ethereumjs?

components/Reference/DocRow.tsx
```diff
@@ -109,7 +109,11 @@ const DocRow = ({
             >
               <MDXRemote {...itemDoc.mdxSource} components={docComponents} />
               {dynamicFeeForkName && dynamicFeeDocMdx && (
-                <MDXRemote {...dynamicFeeDocMdx} components={docComponents} />
+                <MDXRemote
+                  compiledSource={''}
```
assimovt, line 113: nit: no interpolation needed, `compiledSource=""` should do

components/Reference/index.tsx
```diff
@@ -258,7 +262,13 @@ const ReferenceTable = ({
...
                     <td colSpan={colSpan}>
                       <DocRow
-                        itemDoc={itemDocs[opcodeOrAddress]}
+                        itemDoc={
+                          isAfterMerge
+                            ? opcodeOrAddress == '44'
+                              ? itemDocs['44_merge']
+                              : itemDocs[opcodeOrAddress]
+                            : itemDocs[opcodeOrAddress]
+                        }
```
assimovt, line 271: I'd extract this into useMemo method `itemDoc` with if/elses to make it easier to read.

context/ethereumContext.tsx
```diff
@@ -157,7 +159,10 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
       chain: Chain.Mainnet,
       hardfork: fork || CURRENT_FORK,
     })
-    vm = new VM({ common })
+
+    vm = await VM.create({ common })
+
+    evm = await EVM.create({ common, eei: vm.eei })
```
assimovt, line 165: this is interesting that there is no a single evm instance and we have to use both. Is this something coming from ethereumjs?

context/ethereumContext.tsx
```diff
@@ -413,7 +426,7 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
 
     common.hardforks().forEach((fork) => {
       // ignore null block forks
-      if (fork.block) {
+      if (fork.block || fork.name === 'merge') {
```
assimovt, line 429: can we extract the name of the merge fork in a const

context/ethereumContext.tsx
```diff
@@ -430,17 +443,40 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
...
-          opcodeOrAddress: toHex(op.code),
-          staticFee: op.fee,
-          minimumFee: 0,
-          name: op.fullName,
-        },
-      }
+      // TODO: need to implement proper selection of doc according to selected fork (maybe similar to dynamic gas fee)
+      // Hack for "difficulty" -> "prevrandao" replacement for "merge" HF
+      const opcode =
+        selectedFork?.name === 'merge'
```
assimovt, line 451: same

context/ethereumContext.tsx
```diff
@@ -430,17 +443,40 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
...
+          ? toHex(op.code) == '44'
+            ? {
+                ...meta['44_merge'],
```
assimovt, line 454: same

context/ethereumContext.tsx
```diff
@@ -430,17 +443,40 @@ export const EthereumProvider: React.FC<{}> = ({ children }) => {
...
+              }
+          : {
+              ...meta[toHex(op.code)],
+              ...{
+                opcodeOrAddress: toHex(op.code),
+                staticFee: op.fee,
+                minimumFee: 0,
+                name: op.fullName,
+              },
+            }
```
assimovt, line 479: think we'd benefit from extracting selecting a doc into its own func, this piece of code is quite hard to follow otherwise.

util/compiler.ts
```diff
@@ -19,10 +19,7 @@ export const getTargetEvmVersion = (forkName: string | undefined) => {
   if (forkName === 'muirGlacier') {
     return 'berlin'
   }
-  if (forkName === 'arrowGlacier') {
-    return 'london'
-  }
-  if (forkName === 'grayGlacier') {
+  if (forkName && ['arrowGlacier','grayGlacier','merge'].includes(forkName)) {
```
assimovt, line 22: nit: spaces around members of the array

assimovt (changes requested):
hey @thevaizman thx for the PR and for going down the rabbit hole with the ethereumjs/TS issues. I left a few comments, mainly on the code styling, but otherwise LGTM. If there is an existing branch for properly fixing [the issue](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2295), so that we can get rid of the build script, we could already create another issue in this repo to track it and update to the latest. Cheers ✌🏽

prebuild.sh
holgerd77: Hi there, I will prepare the new EthereumJS releases today and releases are planned for tomorrow.

docs/opcodes/44_merge.mdx
holgerd77: Just a short mention: we can't do a short-term update on this due to compatibility issues, so this won't go into the imminent releases. We'll need to give this an additional analysis if it's worth to add an alias (we can't just remove "Merge", again: backwards compatibility). Generally you are totally right, this is mis-named.

prebuild.sh
thevaizman: So the fix had already been merged to their `master` branch. See [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2303).
And from the last comment in the PR, it looks like this will get integrated into `release` immediately, so I'd say we wait for them to integrate it and then I'll push another commit to this branch that will delete the `prebuild.sh` script.

package.json
thevaizman: Hey, you're right, this was just some leftovers I forgot to clean from when I was testing some stuff.
Removing this package.

context/ethereumContext.tsx
thevaizman: So that one is a bit annoying, but I couldn't find a better way around it.
We use `evm`'s `getActiveOpcodes` function in our component. The problem with it, is that since `@ethereumjs`'s latest version had separated the `evm` and `vm` classes, it is no longer possible to access this function using `vm`.
Even trying something like `vm.evm.getActiveOpcodes` isn't working, so I had to resort to instantiating an `evm` object and access the function this way.

context/ethereumContext.tsx
thevaizman: refactored this block of code a little bit and extracted it to a separate function.

components/Reference/index.tsx
thevaizman: Struggling a bit with defining it as a `useMemo` since this function has the `opcodeOrAdderess` dependency.
If you could please suggest a change, I can commit it :)

context/ethereumContext.tsx
holgerd77: Can you eventually open a short issue on this? This is something we should definitely address.

prebuild.sh
holgerd77: Release PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2361

context/ethereumContext.tsx
thevaizman: @holgerd77 sure thing. It's [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2363).

components/Reference/index.tsx
thevaizman: Ah sure, makes sense to use a callback here 👍 
Changed the code locally and I will test and merge everything tomorrow when we hopefully merge this PR.

dorlevi (approved):

dorlevi (approved):

dorlevi (approved):

PR #182, opened 2022-10-16
fix: "call data" to calldata

ethosdev:
Rightly or wrongly, most places on the web use calldata as a single word.

For example, the Intrinsic Gas section of this About page also uses "calldata".


thevaizman (approved):
Thanks, appreciate the contribution :)

PR #183, opened 2022-10-18
EVM: Solve the error related to anchor key redirection on about page

michyaraque:


michyaraque:
This PR contains a change to the about page. The bug fixed is as follows: 
When a user clicks on any link that needs to be redirected to the home page, the link does not work because the home page anchor key needs to have the specified fork query.

peroket:
@michyaraque sorry to get back to you so late, thanks for this change. Would you have the capacity to update it to the latest version of the about page? Then I would be happy to merge it.

czsko707 (approved):

PR #191, opened 2022-11-13
Update 3F.mdx

beultra:
Minor correction of the result specification.
See yellow paper, and geth implementation of extcodehash (https://github.com/ethereum/go-ethereum/blob/8334b5f51a16b72cf2b3ebdc9e131a442c5289d7/core/vm/instructions.go#L392)


peroket (approved):
Thank you :-)

PR #200, opened 2022-12-27
Feat: Live bytecode mode in Editor

vicnaum:
Proposing a live-updating editor in Bytecode mode - it shows the opcodes on the right when writing bytecode.
Should help much when analyzing/writing minimal contracts directly in Bytecode.

Features added:
- Live-update of the opcodes when editing the bytecode
- Ability to put spaces and newlines in bytecode without errors in compilations
- Ability to put comments in bytecode (three styles supported: `// comment`, `# comment`, `; comment`)
- Added `bytecode` language to HLJS with comments highlighting
- Delayed bytecode error validation for hex and even-byte length (for better typing UX)

Maybe something else can be added if interactive mode is worth going forward with, like:
- A switch "Interactive mode: ON/OFF"
- Show decoded opcodes on bytecode page loads
- Interchange between Bytecode/Mnemonics with the same instructions (saw in code, but don't know if it works)
- Add interactive mode for Mnemonics (Not sure about YUL/Solidity - it takes much longer to compile to be interactive)

### DIFF
--- a/components/Editor/index.tsx
+++ b/components/Editor/index.tsx
@@ -285,26 +285,29 @@ const Editor = ({ readOnly = false }: Props) => {
     }
   }
 
+  const stripBytecode = (value: string) => {
+    return value
+      .replaceAll(/\/\/.*$/gm, '')
+      .replaceAll(/;.*$/gm, '')
+      .replaceAll(/#.*$/gm, '')
+      .replaceAll(/\s/gm, '')
+  }
+
   const handleCodeChange = (value: string) => {
     setCode(value)
     setCodeModified(true)
 
     try {
       if (codeType === CodeType.Bytecode) {
-        const concatCode = value
-          .replaceAll(/\/\/.*$/gm, '')
-          .replaceAll(/;.*$/gm, '')
-          .replaceAll(/#.*$/gm, '')
-          .replaceAll(/\s/gm, '')
-        console.log(`|${concatCode}|`)
+        const cleanBytecode = stripBytecode(value)
 
         if (timeOutId) {
           clearTimeout(timeOutId)
           setTimeOutId(undefined)
         }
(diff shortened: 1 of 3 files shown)


components/Editor/index.tsx
```diff
@@ -271,26 +274,61 @@ const Editor = ({ readOnly = false }: Props) => {
...
+      if (codeType === CodeType.Bytecode) {
+        const cleanBytecode = stripBytecode(value)
+
+        if (timeOutId) {
+          clearTimeout(timeOutId)
+          setTimeOutId(undefined)
+        }
+        setTimeOutId(setTimeout(() => validateBytecode(cleanBytecode), 1000))
+
+        loadInstructions(cleanBytecode)
```
darkdarkdragon, line 310: You're calling `loadInstructions` without validating code.
Wouldn't it be better to call inside timeout and after passing validation?

darkdarkdragon (changes requested):

components/Editor/index.tsx
vicnaum: Theoretically yes, but from what I saw in `loadInstructions()` code - if it's invalid - it outputs "INVALID" opcode, which also gives some information about what you've typed is invalid. Otherwise the opcodes view would just not update, which at least for my taste isn't a pleasant flow.
But semantically, if loadInstructions could at some point in future crash the whole application - that might be a problem.
Similar case with calling it inside timeout - it will update with a delay, even if it's correct. Instant refresh is a killer feature here IMHO.

darkdarkdragon (dismissed):

components/Editor/index.tsx
darkdarkdragon: I see, makes sense.

darkdarkdragon (approved):

PR #205, opened 2023-02-07
Add space between words 'our' and 'reference' in About

vegarsti:
Thanks for a very useful website. Just fixing this:

Screenshot 2023-02-07 at 14 32 23

### DIFF
--- a/pages/about.tsx
+++ b/pages/about.tsx
@@ -257,8 +257,7 @@ const AboutPage = () => {
               <b>Opcode Fixed Execution Cost</b> : Each opcode has a fixed cost
               to be paid upon execution, measured in gas. This cost is the same
               for all executions, though this is subject to change in new
-              hardforks. See our
-              {' '}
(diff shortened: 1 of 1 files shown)


pages/about.tsx
```diff
@@ -258,6 +258,7 @@ const AboutPage = () => {
               to be paid upon execution, measured in gas. This cost is the same
               for all executions, though this is subject to change in new
               hardforks. See our
+              {' '}
```
peroket, line 261: I think it has to go on the same line.
```suggestion
              hardforks. See our{' '}
```

peroket (changes requested):
Thank you for finding and fixing that :-)
Just need to fix the syntax to compile. If you notice also there is another one just below with the same issue, would appreciate if you can add a space there too while you're at it.

vegarsti:
Thanks! Like this?

peroket (approved):
Yes thanks a lot :-)

PR #210, opened 2023-02-09
Fix: gas calculation modexp precompile

dragan2234:
Here's a small bugfix for gas calcuation modexp precompile. Resolves https://github.com/comitylabs/evm.codes/issues/209

### DIFF
--- a/util/gas.ts
+++ b/util/gas.ts
@@ -578,12 +578,14 @@ export const calculatePrecompiledDynamicFee = (
       }
 
       if (common.gteHardfork('berlin')) {
-        result.imul(multComplexityEIP2565(maxLen)).divn(Gquaddivisor)
+        result.imul(multComplexityEIP2565(maxLen))
+	 result = result.divn(Gquaddivisor)
         if (result.ltn(200)) {
           result = new BN(200)
         }
       } else {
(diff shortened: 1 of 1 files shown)


util/gas.ts
```diff
@@ -578,12 +578,14 @@ export const calculatePrecompiledDynamicFee = (
       }
 
       if (common.gteHardfork('berlin')) {
-        result.imul(multComplexityEIP2565(maxLen)).divn(Gquaddivisor)
+        result.imul(multComplexityEIP2565(maxLen))
+	result = result.divn(Gquaddivisor)
```
peroket, line 582: How about instead?
```suggestion
        result.imul(multComplexityEIP2565(maxLen)).idivn(Gquaddivisor)
```

peroket (changes requested):
I understand the mistake now, thank you.
There seems to be some alignment issues, hence asking to fix it, but otherwise looks good. You can also just add the missing 'i' instead, like suggested in my comment.

util/gas.ts
dragan2234: thanks @peroket

peroket (approved):

PR #241, opened 2023-05-16
chore: do not persist selected fork, confusing

assimovt:


shaneauerbach:
LGTM. FWIW, I also kinda wished we sorted the forks in the dropdown by descending order of recency (current it's ascending) so that the most recent was at the top. Not sure if you disagree @peroket @thevaizman @assimovt

thevaizman:
> LGTM. FWIW, I also kinda wished we sorted the forks in the dropdown by descending order of recency (current it's ascending) so that the most recent was at the top. Not sure if you disagree @peroket @thevaizman @assimovt 

@shaneauerbach agree it should be desc

peroket (approved):

PR #244, opened 2023-06-05
Feat: editable stack values

jinukaloshaechi:
I propose adding the ability to edit the stack value. This would be helpful when debugging.  
I'm not familiar with typescript, so I implemented it like a PoC.

https://github.com/smlxl/evm.codes/assets/109196262/67aa4944-bd8e-4f89-8be5-aed54110c83c


peroket:
Thank you for your contribution, we will come back to you shortly :-)

dorlevi:
@jinukaloshaechi : Thanks so much for your contribution, especially as you are not too familiar with Typescript. 

We discussed this internally and at this point we won't be taking this feature further, while we think it's an interesting feature and there will be some scenario and users that will benefit from it, we personally believe that taking it from the current state to one that would be shippable would take more effort and perhaps not one that would be worth the investment/added code given it's quite niche and the visibility of this feature will be fairly limited without design changes, examples of some of what we think will need to be done for feature to be "complete"

- Some design changes to make the experience of editing more intuitive/visible
- Adding visual queues that a value was overridden/edited
- Editing of the value not via this modal (which we don't use elsewhere)
- Adding it to storage/memory (not a must but would make sense)
- Permalinks to support the edits (not a must again)
- adding tests, cleaning up code

Let me know if this make sense.

jinukaloshaechi:
@dorlevi 
Sorry for the late reply.

I think this change will help a very small percentage of people (e.g. hacker reversing evm bytecode).
However, I think it would be useful for a lot of people if the ability to modify storage and memory as well as the stack was added.

Unfortunately, I'm not a professional developer, so it would be difficult for me to give design advice for a project like this. I'll settle for using the "editable stack" feature on myself.

Thank you for your interest in my pull-request. It was fun to try out pull-requests on a project that I find useful 😀 

Can I close the PR? Thank you!

peroket:
Thank you for your contribution, I'm closing the PR :-)

PR #248, opened 2023-06-16
fix: wrong `topic_count` value for `log` opcodes

leovct:
I think the `topic_count` values of `log` opcodes are wrong.

Here's a quick example with [`log0`](https://www.evm.codes/#a0?fork=shanghai).
- The minimum gas cost is 375
- The dynamic gas cost is given by the formula `dynamic_gas = 375 * topic_count + 8 * size + memory_expansion_cost`
- It is said that for `log0`,  `topic_count` is equal to 0. I think it should be 1 in this case instead.

Screenshot 2023-06-16 at 10 34 50
Screenshot 2023-06-16 at 10 34 59

### DIFF
--- a/docs/opcodes/A0/homestead.mdx
+++ b/docs/opcodes/A0/homestead.mdx
@@ -3,5 +3,5 @@
     static_gas = {gasPrices|log}
     dynamic_gas = {gasPrices|logTopic} * topic_count + {gasPrices|logData} * size + memory_expansion_cost
 
-For LOG0, `topic_count` is 0.
+For LOG0, `topic_count` is 1.
 The memory expansion cost explanation can be found [here](/about).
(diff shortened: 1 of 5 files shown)


peroket:
Thank you for contributing. 

This is not the case. See for example here https://medium.com/mycrypto/understanding-event-logs-on-the-ethereum-blockchain-f4ae7ba50378
You can also notice that LOG0 has no topic parameters, LOG1 has 1 topic parameter and so on.
Please check your assumptions before opening a PR :-)

Closing the PR.

leovct:
Oh I got it wrong, `gas_cost = static_gas + dynamic_gas`. My bad :(

peroket:
No worries :-)
Please let us know if we can make it more clear.

leovct:
I was thinking of something like the following.

Screenshot 2023-06-16 at 15 54 35

PR #249, opened 2023-06-16
doc: improve dynamic gas computation

leovct:
I opened a first PR (#248) thinking that the gas computation for `log` opcodes was wrong. It turns out I was wrong but, as a newbie in this field, I find that the gas calculation is not 100% clear and could perhaps be improved? At the moment, the gas calculation mentions a static gas and a dynamic gas cost, but there's no formula linking the two (even if it's trivial, I can hear it). That's why I suggest adding the following formula: `gas_cost = static_gas + dynamic_gas` at the top of gas computations, to make it a bit clearer.

Screenshot 2023-06-16 at 16 05 28

On top of that, I transformed some of the text into formulas. I think that it ensures consistency between the different ways of calculating gas. One thing I'm not 100% sure though is how to deal with multi-line formulas such as the one used for the `balance` opcode in the `Berlin` hard fork.

This is how it renders:

Screenshot 2023-06-16 at 16 06 24

Another option may be the following:

Screenshot 2023-06-16 at 16 20 06

### DIFF
--- a/docs/opcodes/0A/homestead.mdx
+++ b/docs/opcodes/0A/homestead.mdx
@@ -1,5 +1,6 @@
 ## Gas
 
+    gas_cost = static_gas + dynamic_gas
     static_gas = {gasPrices|exp}
     dynamic_gas = {gasPrices|expByte} * exponent_byte_size
 
--- a/docs/opcodes/20/homestead.mdx
+++ b/docs/opcodes/20/homestead.mdx
@@ -2,6 +2,7 @@
 
     minimum_word_size = (size + 31) / 32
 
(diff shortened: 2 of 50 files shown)


peroket:
Thank you for your contribution.

We think it may be too much, as the formula is already clearly visible on the right side under the "Estimate your gas cost" (see example image)
image

For the other formula change, we determined that it would probably be clearer to do them as text. Especially if there is both formula and then text to add conditions to it, I fear people will skip the text and miss some part of it.

So I'm closing this PR, and thank you for challenging and making us think how to improve evm.codes :-)

PR #253, opened 2023-06-27
fix: `staticcall` opcode was introduced during the `Byzantium` hard fork

leovct:
For reference: https://blog.ethereum.org/2017/10/12/byzantium-hf-announcement

- https://github.com/ethereum/execution-specs/blob/master/src/ethereum/byzantium/vm/instructions/system.py#L473
- https://github.com/ethereum/execution-specs/blob/master/src/ethereum/spurious_dragon/vm/instructions/system.py#L594

### DIFF
--- a/docs/opcodes/FA.mdx
+++ b/docs/opcodes/FA.mdx
@@ -1,5 +1,5 @@
 ---
-fork: Homestead
+fork: Byzantium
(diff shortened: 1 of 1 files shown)


peroket (approved):
Thank you :-)

PR #257, opened 2023-07-11
Add evm storage banner

shaneauerbach:
I added an evm.storage banner. It can be closed, but it will reopen when the user refreshes the page. My thought is that we would only use this banner for a few weeks and then we could explore a more permanent place to add the link.

Tried to do this in a way that was clean and aligned with @alannafischer's designs. May not be quite pixel perfect but should be very close.

[will fix the prettier stuff now]

### DIFF
--- a/components/EvmStorageBanner.tsx
+++ b/components/EvmStorageBanner.tsx
@@ -1,24 +1,24 @@
-import { useState, useEffect } from 'react'
+import { useState, useEffect } from 'react';
 
 const EvmStorageBanner = () => {
-  const [isShown, setIsShown] = useState(false)
+  const [isShown, setIsShown] = useState(false);
 
   useEffect(() => {
     // Check if the banner was closed previously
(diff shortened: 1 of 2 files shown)


shaneauerbach:
Alright, per @dorlevi's feedback, I made it so that if you ever close the banner it saves that state in localStorage and therefore doesn't open the banner upon reload. Less spammy. The site assumes it should hide it before checking localStorage--I did this because otherwise it would briefly load the banner and then hide it after checking localStorage on reload.

To test:
1. Open [the preview](https://evm-codes-git-add-evm-storage-banner-smlxl.vercel.app). You should see the banner.
2. Close the banner.
3. Reload the page. You should not see the banner.
4. Run `localStorage.clear()` in console in browser dev tools (e.g., inspect).
5. Reload the page. You should see the banner.
6. If you're feeling creative, you could repeat this on mobile. I tried it in an emulator and worked fine.

dorlevi (approved):

PR #266, opened 2023-09-08
Update `ISZERO` description

pcaversaccio:
As title. I think this makes more sense than _Simple not operator_.


peroket (approved):

Issue #267, opened 2023-09-11
`SSTORE` error cases do not include `GAS_CALL_STIPEND` condition


MarkuSchick:
The `SSTORE` opcode requires that the gas left is greater than 2300 (see: [executions spec](https://github.com/ethereum/execution-specs/blob/9de8978795a488a7a2250a5c2c2a2576340b27b1/src/ethereum/istanbul/vm/instructions/storage.py#L71) , [geth](https://github.com/ethereum/go-ethereum/blob/12ef276a7d7faf29e504c52f2efe329a55ab895c/core/vm/gas_table.go#L170)) 
This is important because *transfer* and *send* only pass 2300 gas.

PR #268, opened 2023-09-11
feat: gasleft condition for sstore

MarkuSchick:
fixes https://github.com/smlxl/evm.codes/issues/267


peroket (commented):

Just two notes: 'gasleft' is more of a coding variable name, I would rephrase to something like 'when the amount of gas left to the transaction is less than or equal 2300'. The second point is that this is true only since the Istanbul revision, so we need to mention that as well.

MarkuSchick:
> Istanbul

Thanks for the great feedback. I use evm.codes every day and am happy to contribute :)

Have fixed it here https://github.com/smlxl/evm.codes/pull/268/commits/bd5005f74d8bf88990bf31849bddc5d64c14f69c

peroket (approved):
Thank you very much :-)

PR #271, opened 2023-09-29
Explain how initialisation code is provided to CREATE

PaulRBerg:
The current description does not say anything about how the initialisation code is provided by the end user.

### DIFF
--- a/docs/opcodes/F0.mdx
+++ b/docs/opcodes/F0.mdx
@@ -3,11 +3,11 @@ fork: Frontier
 group: System operations
 ---
 
-*Index 1 is top of the stack. See [PUSH](/#60).*
+_Index 1 is top of the stack. See [PUSH](/#60)._
 
 ## Notes
 
-Creates a new contract. Enters a new sub [context](/about) of the calculated destination address and executes the provided initialisation code, then resumes the current context.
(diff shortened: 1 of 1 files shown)


docs/opcodes/F0.mdx
```diff
@@ -7,7 +7,7 @@ group: System operations
 
 ## Notes
 
-Creates a new contract. Enters a new sub [context](/about) of the calculated destination address and executes the provided initialisation code, then resumes the current context.
+Creates a new contract using the initialisation code provided as [calldata](/about). Enters a new sub [context](/about) of the calculated destination address and executes the initialisation code, then resumes the current context.
```
peroket, line 10: I would disagree here, this is not calldata at all, it's the actual code to the constructor. Which is explained in the input section.

peroket (commented):

However, I think this is not helping, and even misleading.

docs/opcodes/F0.mdx
PaulRBerg: Good point, but I still feel like an explanation is needed. How about:

> Creates a new contract using the initialisation code provided in the data field of the transaction

peroket (approved):
Thank you :-)

PR #272, opened 2023-10-06
chore: Clean up chain ids

ardislu:
This PR deletes a broken link, removes deprecated testnets, and adds the chain ID for the new [Holešky testnet](https://github.com/eth-clients/holesky).


dorlevi (approved):
thanks for the update!

PR #278, opened 2023-12-03
Focus Opcode-Search on `Alt+K`

### LINKED ISSUE #134: Feature request: Make it easy to to move between opcodes
## Motivation

I know there is a search bar on top right, but while I'm viewing some opcode and have to see another one, it needs to be scrolled all the way to the top and click on the search bar. Doing this one or twice might not seem a problem, but imagine writing a bytecode and referring to evm.opcodes and moving through opcodes 100s of times every time need to scroll. I try to get away with hitting Command + F (or Ctrl + F), however, the browser search is terrible it matches so many occurrences that it's more annoying to reach where I want.

## Feature Request

I believe any one of the following can make browsing opcodes fast. (Or if I missed an existing fast way to browse opcodes pls lmk).

- When Command + F is hit, it would be super awesome if the focus moves to the search bar.
- Or if the opcodes are added to kbar, e.g. if I do command + K and type "STATI..." it should show STATICCALL in the dropdown and I can just hit return and it takes the view to the right opcode in the opcodes list.

## Acknowledgements

Really helpful site! If you guys would like, I can make a PR for the above feature, pls lmk what you think.

lmittmann:
This PR adds the keybinding `Alt+K` to focus the opcode search input (resolves #134).

### DIFF
--- a/components/Reference/Filters.tsx
+++ b/components/Reference/Filters.tsx
@@ -48,7 +48,7 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
   }
 
   const handleAltK = (event: KeyboardEvent) => {
-    if (event.altKey && (event.key === 'k' || event.key === 'K')) {
+    if (event.altKey && (event.key.toLowerCase() === 'k')) {
       inputRef.current?.focus()
       inputRef.current?.value && inputRef.current.select()
     }
--- a/components/Reference/Filters.tsx
+++ b/components/Reference/Filters.tsx
@@ -1,4 +1,4 @@
-import { useState, useMemo, useEffect } from 'react'
+import { useState, useMemo, useEffect, useRef } from 'react'
 
 import debounce from 'lodash.debounce'
 import { useRouter } from 'next/router'
@@ -33,6 +33,8 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
     [isPrecompiled],
   )
 
+  const inputRef = useRef<HTMLInputElement>(null)
+
   const handleKeywordChange = debounce(
     (value: string) => onSetFilter(searchFilter.value, value),
     debounceTimeout,
@@ -43,6 +45,15 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
     onSetFilter(searchFilter.value, '')
     setSearchKeyword('')
     setSearchFilter(option)
+  }
+
+  const handleCtrlF = (event: KeyboardEvent) => {
+    if (event.ctrlKey && event.key === 'f') {
+      event.preventDefault()
+
+      inputRef.current?.focus()
+      inputRef.current?.value && inputRef.current.select()
+    }
   }
 
   // Change filter and search opcode according to query param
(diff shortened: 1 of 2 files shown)


peroket:
Thank you, @lmittmann, for your contribution. However, this change did not really reach consensus in our company, and we would prefer not to apply it. We find that the behaviour is not really intuitive and would be confusing rather than helping.

lmittmann:
@peroket do you find the key-binding confusing, or a key-binding to jump to the opcode-search in general?

shaneauerbach:
> @peroket do you find the key-binding confusing, or a key-binding to jump to the opcode-search in general?

Hey @lmittmann, that's a good question and something I was talking about with @peroket. Our initial concern was that we felt overwriting expected browser behavior could be confusing to the user, and that conditioning behavior on the focus might also feel counterintuitive--it makes some sense once you know it, but less so if you don't. 

This is a little less problematic if we're not using CMD/CTRL-F, and @peroket mentioned that he generally he prefers OPT/ALT+ key-binds on sites to avoid collisions with the browser. But if we used a different key-bind, how would we communicate it to the user?

These are the questions that had us a bit stumped and made us conclude it wasn't worth pursuing. If you have a good plan in mind that would address these, please do share. We really appreciate your contribution, and it was implemented pretty cleanly.

To turn the question around a little: how would this feature improve your experience? I often use the browser's CMD/CTRL-F on evm.codes. It doesn't use the filter, but it does generally take me where I want to go. Seems like this doesn't serve your use case as well--please share more on that if you can.

shaneauerbach:
Thanks for further explaining, @lmittmann. I agree it would be nice to have.

We're still uncomfortable with the conditional logic around CTRL-F. We'd prefer to simply avoid collisions with browsers. I'm fine with how we're using CTRL-K currently because that's a convention that many sites use, but generally we'd prefer to use ALT for other hotkeys. ALT-F unfortunately collides with some browser/OS combos. Would you be open to using ALT-K as the hot key for the keyword search? Then you have it all to yourself and we don't need the logic to accommodate both the browser's and our functionality with the single key.

I also like your idea of communicating the ALT-K hotkey in the placeholder text within the search input.

Thoughts?

lmittmann:
@shaneauerbach can you reopen the PR? I updated the keybinding to `Alt+K` and added a visual indication for it.

shaneauerbach:
Thanks, @lmittmann. @mcmxcdev could you please provide a code review? I've assigned you.

components/Reference/Filters.tsx
```diff
@@ -45,6 +47,13 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
     setSearchFilter(option)
   }
 
+  const handleAltK = (event: KeyboardEvent) => {
+    if (event.altKey && (event.key === 'k' || event.key === 'K')) {
```
mcmxcdev, line 51: This should work too and is more concise.

```suggestion
    if (event.altKey && (event.key.toLowerCase() === 'k')) {
```

components/Reference/Filters.tsx
```diff
@@ -56,6 +65,11 @@ const Filters = ({ onSetFilter, isPrecompiled = false }: Props) => {
       handleKeywordChange(query.name as string)
       router.push(router)
     }
+
+    // Register and clean up Ctrl+F event listener
```
mcmxcdev, line 69: This would now be Alt+K ;)

mcmxcdev (approved):
Great work thanks!

I left two optional comments but they don't influence behavior, so good to merge whenever.

mcmxcdev:
PR description needs an update (still references `Ctrl+F`)

lmittmann:
@mcmxcdev thanks for the review. I applied your suggestion and updated the PR description.

mcmxcdev (approved):

zemse:
This is great! Thanks @lmittmann for all the work!

I am on a mac and I don't have `Alt` key, I tried using the `option` key however it doesn't focus there. Can any mac user try this please?

PR #291, opened 2024-01-19
Fix issue with backslash crashing the app

### LINKED ISSUE #290: App crashes after inputting '\' into Enter keyword input
https://github.com/smlxl/evm.codes/assets/15238391/17b5d8e4-f044-43fb-ac6d-d0736e2c955f

2xic:
Fixes #290 

## Test plan


components/Reference/columns.tsx
```diff
@@ -2,9 +2,13 @@ import { Row } from 'react-table'
 
 import { StackBox } from 'components/ui'
 
-const filter = (rows: Row[], id: string, filterValue: string) => {
-  const re = new RegExp(`${filterValue}`, 'i')
```
2xic, line 6: This is just a case insenctive matching pattern - instead of using regex which can cause issues when users write invalid patterns, we use vanilla javascript 😄

components/Reference/columns.tsx
```diff
@@ -2,9 +2,13 @@ import { Row } from 'react-table'
...
-  return rows.filter((row: any) => row.original[id].match(re))
+// Possible fields are defined in `Opcodes.json`
+type OpcodeRow = Row<Record<string, string | undefined>>
```
2xic, line 6: Not able to get this fully typed as it depends on the json file, but this is better than the `any` type

mcmxcdev (approved):

PR #294, opened 2024-01-30
fix: add changes from eip3651

trocher:
This PR includes the following changes to the `about` page:

- add details about [EIP3651](https://eips.ethereum.org/EIPS/eip-3651): the `COINBASE` address is always warm 
- add missing details about [EIP2929](https://eips.ethereum.org/EIPS/eip-2929): `tx.sender` and `tx.to` are always warm
- fix some typos


thevaizman (approved):
@trocher thanks for your contribution!

PR #297, opened 2024-02-08
Contract Viewer

hananbeer:
The goal is to add a Contract Viewer tab with better functionalities than etherscan, ideally independent of etherscan and open source.

See demo video here:
https://x.com/high_byte/status/1743294802022609172?s=20

Things I've added:

ui components live in /components/ContractViewer.ts. I use shadcn components which live in the /@ folder, mui-tree-viewer element, and monaco editor. I think most of the CR should focus here since I'm no react developer.

api endpoint in /pages/api/getContract.ts to fetch etherscan contract information

utilities to flatten solidity source code (with small caveat of not yet supporting named imports) in /util/flatten.ts

utilities to parse etherscan info in /util/EtherscanApi.ts and /util/EtherscanParser.ts (in the end we want to support sourcify too)

reworked the solidity compiler web worker - supports selecting compiler version & generally just nicer smaller code.

I ended up using abi & storageLayout from solc, not using ast anymore at this moment.


pages/about.tsx
2xic, line 1: Can we move the file changes in this file to a separate PR as they aren't needed for the contract viewer feature ?

util/solc.ts
```diff
@@ -0,0 +1,99 @@
...
+  worker?: Worker
+  callbacks: { [id: string]: (data: any) => void } = {}
+
+  // NOTE: this lazy-load instead of constructor is due to a bug:
+  // "ReferenceError: Worker is not defined"
+  // this lazy-loading seems to fix it
+  init() {
+    if (!this.worker) {
+      // console.info('starting solc worker')
+      this.worker = new Worker('/solcWorker.js')
```
2xic, line 13: Why is this file used ? 

You added `solcworker2` so I assumed it would be used instead of the old `solcWorker.js`

babel/solcWorker2.js
```diff
@@ -0,0 +1,161 @@
+import wrapper from 'solc/wrapper'
```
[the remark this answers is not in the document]
2xic: I also don't see this file used anywhere - is it in use ? 🤔

components/ContractViewer/AstProcessor.ts
```diff
@@ -0,0 +1,92 @@
...
+  node: T
+  // undefined for global scope
+  scope?: AstTypes.ContractDefinition | AstTypes.FunctionDefinition
+}
+
+// this represents a single contract/interface/library definition block
+export class SourceDefinition {
+  // name: string = ''
+  // filepath: string = ''
+  // c3 inheritance / ast node ref?
```
2xic, line 17: Can you clarify or remove the comments here ?

components/ContractViewer/ContractTreeNode.tsx
```diff
@@ -0,0 +1,661 @@
...
+  }
+
+  return (
+    <TextField
+      size="small"
+      readOnly={output}
+      error={error}
+      sx={{ marginTop: '4px', marginBottom: '4px', marginRight: '16px' }}
+      onChange={(e: any) => {
+        // console.log(`updating ${path)}`)
```
2xic, line 229:
```suggestion
```

components/ContractViewer/ContractTreeNode.tsx
```diff
@@ -0,0 +1,661 @@
...
+
+  if (inputAbi.type == 'tuple') {
+    return (
+      <div className="border rounded-xl pl-2 py-2 mr-2 hover:border-blue-500">
+        <TupleParamItem path={path} inputAbi={inputAbi} reducer={reducer} />
+      </div>
+    )
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
2xic, line 197: I think this can be removed also 
```suggestion
```

components/ContractViewer/ContractTreeNode.tsx
```diff
@@ -0,0 +1,661 @@
...
+
+type StorageLayoutItemProps = {
+  id: string
+  address: string
+  storage: {
+    astId: number
+    label: string
+    contract: string
+    offset: number
+    slot: any // string | number
```
2xic, line 486: If it can just be string or number, why have it as a `any` ? 🤔

components/ContractViewer/ContractTreeNode.tsx
```diff
@@ -0,0 +1,661 @@
...
+
+export const StorageLayoutItem = ({ id, address, storage, types }: StorageLayoutItemProps) => {
+  const [status, setStatus] = useState('0x...')
+  const [inputs, setInputs] = useState([])
+
+  const type = types[storage.type]
+
+  // if storage item is a mapping, keyTypes is an array of types of all mapping keys
+  // eg. keys of mapping(address => mapping(uint256 => bytes)) would be ['address', 'uint256']
+  const keyTypes = [...(storage.type.matchAll(/\bt_mapping\((?<key>.+?),/g) || [])].map((m) => m.groups.key)
```
2xic, line 508: I guess using the regex should work just fine, but you could also check if the encoding is a mapping and then read of the key value there.

components/ContractViewer/ContractTreeView.tsx
```diff
@@ -0,0 +1,131 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
2xic, line 1: Ideally this is removed
```suggestion
```

components/ContractViewer/DeploymentInfo.ts
```diff
@@ -0,0 +1,357 @@
...
+
+// TODO: should probably move some of the contract code parts to another class
+// it might be useful to have separate compilationInfo for separate deployments but
+// this class is too big rn, can and probably should move code, abi, compilationInfo to
+// a "SolidityContract" class
+
+/**
+ * DeploymentInfo represents an on-chain deployment:
+ * 
+ * @property chainId (chain id as decimal number)
```
2xic, line 32: Why suddenly start to use this format ? 🤔 

I don't have anything against it per se, but I feel like we should be consistent (we currently don't use this `@property` anywhere else)

components/ContractViewer/DeploymentInfo.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    setDeployments({ ...deployments })
+    if (selectedDeployment == deployment) {
+      setSelectedDeployment(undefined)
+    }
+  }
+
+  return {
+    deployments,
+    // setDeployments,
```
2xic, line 347: How about ? 

```suggestion
```

components/ContractViewer/DeploymentInfo.ts
```diff
@@ -0,0 +1,357 @@
...
+        setSelectedDeployment(deployment)
+
+        if (loadImplementation) {
+          const impl = deployment.etherscanInfo?.Implementation as string
+          if (impl) {
+            /*await*/ loadDeployment(impl.toLowerCase(), deployment)
+          }
+        }
+
+        const selectedOutputs = [/*'ast', 'abi',*/ 'storageLayout']
```
2xic, line 302:
```suggestion
        const selectedOutputs = ['storageLayout']

```

components/ContractViewer/ViewerUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+      ' => ' +
+      getTypePrettyName(type.valueType) +
+      ')'
+    )
+  }
+
+  return type?.name || type?.namePath || ''
+}
+
+export function spaceBetween(str: string, pad = 16) {
```
2xic, line 53: Maybe we should have a test for this function ? 

It's a bit complex so having some test would make it easier to simplify in the future

package.json
```diff
@@ -16,6 +17,8 @@
     "@types/react-dom": "17.0.2"
   },
   "dependencies": {
```
2xic, line 19: We should try to limit the amount of new packages

With this PR we have a increase in especially UI libraries which I think we should be able to reduce.
- emotion
- @mui
- @radix-ui
- tailwind
- _in addition what we already got some ui libraries_

In addition
- `@uidotdev/usehooks` I don't see used anywhere, could it be removed ? 
- `babelify` is used for solcworker2, but that isn't used anywhere, could it be removed ? 
    -  I believe the same is true for `path-browserify`
- I don't see `react-hook-form` used anywhere 
- I don't see `jsx-runtime` used anywhere

2xic (commented):
Thanks for working on this! Very exciting 😄  

----

Left some high level feedback - the build is also failing and I got build error when I tried to run/build it locally (see image) . Can you look into that ?  

Screenshot 2024-02-10 at 21 25 45


**To summarize some of the feedback**

1. We need a deployment build in vercel (and build should work locally also) so that we can test the contract viewer
2. Let's try to minimize the number of new dependencies and aim for consistency where we can
3. We want to have typechecking etc. enabled for the builds 
4. Fork selector must still work
5. Commented out code / comments without context should ideally be rewoerded or removed

.eslintignore
```diff
@@ -1,3 +1,5 @@
 ethereum.js
 solcWorker.js
 public
+@/
```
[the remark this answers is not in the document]
hananbeer: this is not my code. this is shadcn, it installs code directly in the repo instead of node_modules.

babel/solcWorker2.js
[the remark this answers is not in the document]
hananbeer: the files inside `babel` are built independently using `yarn build:solcjs-worker` and loaded via a web worker in `./util/solc.ts`. this is how web dev works apparently, I'm just the messenger.

as per your question about bundle - you already used solc yourself in the playground. if you're referring to the versions list - it's really tiny... 6kb, only on demand.
(the main opcodes page is over 100 times that size and next keeps complaining about that, so priorities)

components/ChainSelector.tsx
```diff
@@ -46,18 +47,23 @@ const ChainSelector = () => {
...
     [onForkChange],
   )
 
-  useEffect(() => {
-    if (defaultForkOption) {
-      handleForkChange(defaultForkOption)
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [defaultForkOption])
+  // TODO: don't redirect (or at least don't override address query param)
```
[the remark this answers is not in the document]
hananbeer: yes, as I mentioned last time I did not intend to commit this.
this is a temp fix because this part breaks my own router. it keeps redirecting and removing my parameters.
instead of touching your code right now I temporarily disabled this (only on load, the chain selector still works)

eventually we will need to fix it there too.

components/ContractViewer/ContractCodeEditor.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import Editor from '@monaco-editor/react'
+import { useTheme } from 'next-themes'
+
+const ContractCodeEditor = ({ code, line, column, ...props }: any) => {
+  const [codeEditor, setCodeEditor] = useState<any>(null)
+  const { theme } = useTheme()
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  function handleEditorDidMount(editor: any, _monaco: any) {
```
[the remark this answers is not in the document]
hananbeer: I thought using `function` would be more clear but since then I realized react doesn't play nicely since `this` isn't preserved... I can change this

components/ContractViewer/DeploymentInfo.ts
```diff
@@ -0,0 +1,357 @@
...
+        }
+      }
+
+      return { result, error }
+    })
+  }
+  
+  // TODO: this is supposed to simply make every function and storage public
+  // but there can be name clashes (eg. two private "name" variables in
+  // different contracts is ok but if both are public then it is an error; need
```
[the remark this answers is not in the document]
hananbeer: this battle will be fought on another merge.

next.config.js
```diff
@@ -1,13 +1,21 @@
...
-  reactStrictMode: true,
+  reactStrictMode: false,
   serverRuntimeConfig: {
     APP_ROOT: __dirname,
   },
+  compiler: {
+    styledComponents: false,
+  },
+  typescript: {
+    ignoreBuildErrors: true,
```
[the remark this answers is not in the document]
hananbeer: unfortunately it won't work, I think due to shadcn. although we can remove shadcn for all I care.

next.config.js
```diff
@@ -1,13 +1,21 @@
...
 const { withPlausibleProxy } = require('next-plausible')
 
 module.exports = withPlausibleProxy()({
-  reactStrictMode: true,
+  reactStrictMode: false,
   serverRuntimeConfig: {
     APP_ROOT: __dirname,
   },
+  compiler: {
+    styledComponents: false,
```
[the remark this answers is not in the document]
hananbeer: I really don't know what that is and how it got there. probably shadcn again.

pages/about.tsx
hananbeer: removed.

util/solc.ts
hananbeer: I compile solcWorker2.js to solcWorker.js. I know it's dumb and confusing, but you used the same filename so I wanted to keep it, but also not override your original work. compatibility.

components/ContractViewer/AstProcessor.ts
hananbeer: if you have a better location for declaration of intents I am open to hear.
otherwise over time these plans will be forgotten.

components/ContractViewer/ContractTreeNode.tsx
hananbeer: the other `any`s did not bother you?

components/ContractViewer/DeploymentInfo.ts
hananbeer: I copied it from somewhere because I thought I would get nice hints on vs code.
as you see I haven't ran with this format any longer since.

components/ContractViewer/DeploymentInfo.ts
hananbeer: yes.

components/ContractViewer/DeploymentInfo.ts
hananbeer: ok.

package.json
hananbeer: usehooks, react-hook-form can be removed. idk about jsx-runtime.
path-browserify was important for solc, but I think I resolved it eventually. will try to remove.
babelify is used to babelify the babel directory.

babel/solcWorker2.js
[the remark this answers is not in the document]
2xic: Oh - I see now, didn't notice it overwrote the current `solcWorker.js`. 

Correct, but doesn't the current logic make it download the same solc twice ? First with the solc package dependency and another time through the `importScripts` ?

.eslintignore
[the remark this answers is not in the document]
2xic: hm - okay if it's some autogenerated code then it should be fine to ignore, but from the [commit history](https://github.com/hananbeer/evm.codes/tree/c84d117ed16f418c99299e8b7cddcf4ffd1e615b/%40/lib) it looked like it was a directory with handwritten component code

next.config.js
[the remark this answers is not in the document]
2xic: okay - I mean if we can replace `shadcn` and then have `ignoreBuildErrors: false,` that would be nice

components/ContractViewer/AstProcessor.ts
2xic: I just meant that I didn't fully understand the comment so was wondering if you could add some more context.

Is the comment meant to say that you want to keep track of the inheritance of the contract with the name and filepath ? 
(if so could you expand upon why that is useful - just so others can follow the thought process)

components/ContractViewer/ContractTreeNode.tsx
2xic: Ideally we don't want `any` if we can have a specific type.

This part of what I was refereeing to by us having typechecking.

components/ContractViewer/DeploymentInfo.ts
2xic: hm okay - it's not that big of a deal, just a bit strange that we just have one place where we use this comment format.

babel/solcWorker2.js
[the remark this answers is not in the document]
2xic: Okay - looked in the preview build and it seems fine.

2xic:
Thanks for fixing the build @hananbeer (it also works for me locally now) 😄 Stability also seems a lot better with this versions so that's great 🎉 

**Some things I noticed:**

We have this loading thing at the bottom when a contract is loaded - is this meant to be included ? (I think it would be nice to have, but then it should likely be higher up on the page)

Screenshot 2024-02-11 at 21 15 14

-------

When pressing `send` on the any of the functions it seems to result in a error and nothing happens 🤔 

-----

Before on function click it would find the function inside the file - is this feature removed now ?

.eslintignore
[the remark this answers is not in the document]
hananbeer: I removed shadcn. marking resolved.

components/ContractViewer/AstProcessor.ts
hananbeer: ok I will remove the comments.

components/ContractViewer/ContractTreeNode.tsx
hananbeer: removed the type comment

components/ContractViewer/DeploymentInfo.ts
hananbeer: feel free to offer a standard and I will use it everywhere.

components/ContractViewer/ViewerUtils.ts
hananbeer: that is a conversation for another day...

package.json
hananbeer: I removed everything I could and restored original version of packages I upgraded. I guess progress is a slow process.

next.config.js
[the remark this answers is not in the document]
hananbeer: it seems to fail on code that isn't mine. idk why.

components/ChainSelector.tsx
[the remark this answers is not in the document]
hananbeer: so since this doesn't break anything, it only fixes things, I will mark it as resolved.
we can discuss this chunk of code on another occasion.

hananbeer:
1. fixed progress bar
2. send seems to work for me. which wallet are you using?
3. yes this feature is removed because it required parsing the ast instead of abi. now it would be ambiguous to do that so I just don't. need to search with ctrl+f and figure out which is the correct function.

2xic (commented):
biggest issue is that `ignoreBuildErrors` is still `true`, but I'll try to help you with it.

After that is fixed, I think we can merge this

charisra:
Superseded by https://github.com/smlxl/evm.codes/pull/318

PR #310, opened 2024-03-31
fix: keccak256 playground example

mtlynch:
The existing playground example uses the SHA3 opcode, which the playground doesn't recognize.

The fix is to replace SHA3 with KECCAK256.

Related: #311


2xic (approved):
Thank you!

PR #311, opened 2024-03-31
fix: replace sha3 with keccak256 in gas calculations

mtlynch:
This fixes a bug where opcodes that relied on the value of sha3 or sha3Word were rendering on the page with a value of 0, giving an incorrect representation of gas fees for the respective opcodes.

This bug appears to be caused by this change in EthereumJS:

https://github.com/ethereumjs/ethereumjs-monorepo/pull/2706/files#diff-aa85288d3ff116040e9adfce5d7d74b582eee859ea89af6cf69cb29f5bccc079

The fix is to switch to the new keccak semantics instead of the deprecated SHA3 semantics.

### Before

image

### After

image

Related #310

### DIFF
--- a/docs/opcodes/20/homestead.mdx
+++ b/docs/opcodes/20/homestead.mdx
@@ -2,7 +2,7 @@
 
     minimum_word_size = (size + 31) / 32
 
-    static_gas = {gasPrices|sha3}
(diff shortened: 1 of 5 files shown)


2xic (approved):
Thank you!!!

Bonus points for a great PR description 🎉

PR #319, opened 2024-04-10
chore: add 0x0a precompile

cre-mer:
I didn't add an example section or a "Reproduce in playground" in playground because of this precompile's multi-layer dependency.


thevaizman (approved):
Thanks for your contribution @cre-mer! 🙏 
For some reason we overlooked this doc for `0x0a` when we updated the website, so thanks for fixing :)

cre-mer:
happy to help 🤗

PR #321, opened 2024-04-12
chore: project maintenance

mcmxcdev:
## Changes
- add `pnpm-lock.yaml`
- remove `yarn.lock`
- add `pnpm-lock.yaml` to `.prettierignore`
- install missing `abitype` dependency
- move from `overrides` in `package.json` to `pnpm.overrides`
- add `packageManager`, `engines` and `volta` fields to `package.json`
- add reported deprecated dependencies to `pnpm.allowedDeprecatedVersions`
- bump up minimum required node version from 18 to 20
- update `README.md` to reference `pnpm` instead of `yarn`
- update `pre-commit` to execute scripts with `pnpm` instead of `yarn`
- update comment in `solcWorker2.js` to reference `pnpm` instead of `yarn`

## Motivation
yarn v1 is deprecated and has had its development frozen since at least 2020, we decided to switch to pnpm.

There are some additional changes in this PR which ended up being necessary to support a smooth switch of package manager.


README.md
```diff
@@ -14,8 +14,8 @@ evm.codes is brought to you by [smlXL](https://www.smlxl.io), powered by open-so
 
 The app requires the following dependencies:
 
-- [NodeJS](https://nodejs.org/) >= 18
-- [Yarn](https://yarnpkg.com/)
+- [NodeJS](https://nodejs.org/) >= 20
```
mcmxcdev, line 17: Node 20 is the active LTS version currently and it aligns this project with other smlxl projects.

package.json
```diff
@@ -5,18 +5,14 @@
...
     "dev": "next dev",
     "format": "prettier --write  --cache .",
     "lint": "eslint --ext js,jsx,ts,tsx .",
     "lint:package": "sort-package-json",
     "typecheck": "tsc --pretty"
   },
-  "resolutions": {
-    "@types/react": "17.0.2",
-    "@types/react-dom": "17.0.2"
-  },
```
mcmxcdev, line 19: `resolutions` is a yarn v1 specific feature (that pnpm can read and understand) but I moved it to `pnpm.overrides` to be more explicit.

package.json
```diff
@@ -100,5 +97,25 @@
...
+  },
+  "packageManager": "pnpm@8.15.6",
+  "engines": {
+    "node": "^20.0.0",
+    "pnpm": "^8.15.6"
+  },
+  "volta": {
+    "node": "20.12.2",
+    "pnpm": "8.15.6"
+  },
```
mcmxcdev, line 109: It is good practice to define these fields to make it clear which version of tooling is supported.

package.json
```diff
@@ -100,5 +97,25 @@
...
+  "pnpm": {
+    "overrides": {
+      "@types/react": "17.0.2",
+      "@types/react-dom": "17.0.2"
+    },
+    "allowedDeprecatedVersions": {
+      "@babel/plugin-proposal-object-rest-spread": "7.12.1",
+      "core-js": "2.6.12",
+      "trim": "0.0.1"
+    }
```
mcmxcdev, line 119: Added these to silence deprecation warnings when running `pnpm install`, but these should be fixed by upgrading respective dependencies using these.

package.json
```diff
@@ -100,5 +97,25 @@
...
+  },
+  "volta": {
+    "node": "20.12.2",
+    "pnpm": "8.15.6"
+  },
+  "pnpm": {
+    "overrides": {
+      "@types/react": "17.0.2",
+      "@types/react-dom": "17.0.2"
+    },
```
mcmxcdev, line 114: In a separate PR, we should look into upgrading `react` and `react-dom` to latest, I don't see a reason why it would be locked to v17.

charisra (approved):
Finally, pnpm !

PR #328, opened 2024-05-26
fix: typo in precompiled/0x06.mdx

vegetabledogdog:


peroket (commented):
There is no typo here. The number are indices, starting from 0, and the range is inclusive. Thus, the 128th byte has index 127.
If you notice, the other end ranges are the same (31, 63, 95).

vegetabledogdog:
> There is no typo here. The number are indices, starting from 0, and the range is inclusive. Thus, the 128th byte has index 127. If you notice, the other end ranges are the same (31, 63, 95).

So, I changed it to 127.

peroket:
> > There is no typo here. The number are indices, starting from 0, and the range is inclusive. Thus, the 128th byte has index 127. If you notice, the other end ranges are the same (31, 63, 95).
> 
> So, I changed it to 127.

Indeed, sorry, looks like I got confused!  Thanks for the fix :-)

peroket (approved):

PR #333, opened 2024-06-11
propagate SoliditySettings when compiling contracts

ilamanov:
When you try to view `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD ` in the Contract Viewer, you get the following compilation errors:
```
[
    {
        "component": "general",
        "errorCode": "6275",
        "formattedMessage": "ParserError: Source \"@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol\" not found: File import callback not supported\n --> contracts/base/Callbacks.sol:4:1:\n  |\n4 | import {IERC721Receiver} from '@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol';\n  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n",
        "message": "Source \"@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol\" not found: File import callback not supported",
        "severity": "error",
        "sourceLocation": {
            "end": 160,
            "file": "contracts/base/Callbacks.sol",
            "start": 71
        },
        "type": "ParserError"
    },
    ...
]
```

Which leads to frontend errors which you can see in the console: 

### The fix
The reason for the above compilation error is that we don't pass in the `remappings` that we get from Etherscan. We need to pass them (and other settings to avoid potentially other errors in the future).

### Test plan
• Try to view `0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD`
• You shouldn't get any errors like in the screenshot above

### DIFF
--- a/components/ContractViewer/DeploymentInfo.ts
+++ b/components/ContractViewer/DeploymentInfo.ts
@@ -8,6 +8,7 @@ import { SourceDefinition } from 'types/ast'
 import {
   EtherscanContractResponse,
   SolidityCompilerOutput,
+  SoliditySettings,
 } from 'types/contract'
 
 import { findContract, flattenCode } from 'util/flatten'
@@ -149,13 +150,35 @@ export class DeploymentInfo {
     }
   }
 
-  async compile(outputs: string[], version?: string) {
+  async compile(
+    outputs: string[],
+    settings?: SoliditySettings,
+    version?: string,
+  ) {
+    if (!settings) {
+      settings = {
+        outputSelection: {
+          '*': {
+            '': [],
+            '*': [],
+          },
+        },
+      }
+    }
+
+    settings.outputSelection['*'][''] = (
+      settings.outputSelection['*'][''] || []
+    ).concat(outputs)
+    settings.outputSelection['*']['*'] = (
+      settings.outputSelection['*']['*'] || []
+    ).concat(outputs)
+
     return solidityCompiler
       .compileCode(
         this.etherscanInfo.SourceCode,
         // if not specifying exact version,
         version || this.etherscanInfo.CompilerVersion,
-        outputs,
+        settings,
       )
       .then(({ result, error }: any) => {
         if (error || !result) {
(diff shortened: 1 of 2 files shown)


components/ContractViewer/DeploymentInfo.ts
```diff
@@ -149,13 +150,35 @@ export class DeploymentInfo {
...
+        },
+      }
+    }
+
+    settings.outputSelection['*'][''] = (
+      settings.outputSelection['*'][''] || []
+    ).concat(outputs)
+    settings.outputSelection['*']['*'] = (
+      settings.outputSelection['*']['*'] || []
+    ).concat(outputs)
```
2xic, line 174: nit, but shouldn't this be part of a else statement ? 

In the case there are no settings, we will create a `outputSelection` with `output` and then concat it again. (this doesn't break anything, but made me have read the code twice to make sure I understood it) 

```suggestion
else {
    settings.outputSelection['*'][''] = (
      settings.outputSelection['*'][''] || []
    ).concat(outputs)
    settings.outputSelection['*']['*'] = (
      settings.outputSelection['*']['*'] || []
    ).concat(outputs)
}
```

2xic (dismissed):
nice 🙌

components/ContractViewer/DeploymentInfo.ts
ilamanov: Oh my bad, small typo. I meant to initialize `outputSelection` to empty arrays instead of to `output`. In that case it doesn't need to be part of the `else` statement.

2xic (approved):
nice 🙌

ilamanov:
### Merge activity

* **Jun 12, 6:40 AM EDT**: @ilamanov merged this pull request with [Graphite](https://app.graphite.dev/github/pr/smlxl/evm.codes/333).

PR #335, opened 2024-06-21
fix: `SSTORE` gas

### LINKED ISSUE #312: SSTORE behaviour typo or wrong description
In the description of `SSTORE` the following code snippet it's displayed as the behavior for the computation of the `base_dynamic_gas`:

```python
static_gas = 0

if value == current_value
    if key is warm
        base_dynamic_gas = 100
    else
        base_dynamic_gas = 100
else if current_value == original_value
    if original_value == 0
        base_dynamic_gas = 20000
    else
        base_dynamic_gas = 2900
else
    base_dynamic_gas = 100
```

Specially in the first **if-else** condition, the assignment in **else** condition seems to be incorrect, given that it's given the same base as the **if** condition, assigning `base_dynamic_gas = 100`. 
```python
if value == current_value
    if key is warm
        base_dynamic_gas = 100
    else
        base_dynamic_gas = 100
```
If this it's correct, then there's no point on checking that it's warm, just assigning `base_dynamic_gas = 100`. I think that maybe this is a typo, and in the else condition, maybe should go `base_dynamic_gas = 2100`, as it's described later.

trocher:
1. Fix an incorrect and unreachable case of the gas refund computation for `SSTORE`
2.  Fix #312 

## 1.

If and only if the following conditions are met:

```
value != current_value
value == original_value != 0
key is cold
```

The gas refund computation would do:

```
gas_refunds += 4900
```

However, this behaviour is not defined by [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200) and [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) and the conditions are impossible to reach given that if `original_value != current_value`, it must be that the key is warm.

## 2.

According to both [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200) and [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929), the defined computations are correct and there is no distinct case when the key is cold or warm. There is hence no need for branches.

### DIFF
--- a/docs/opcodes/55/berlin.mdx
+++ b/docs/opcodes/55/berlin.mdx
@@ -9,17 +9,14 @@
     static_gas = {gasPrices|sstore}
 
     if value == current_value
-        if key is warm
-            base_dynamic_gas = {gasPrices|warmstorageread}
-        else
-            base_dynamic_gas = {gasPrices|sstoreNoopGasEIP2200}
+        base_dynamic_gas = {gasPrices|warmstorageread}
     else if current_value == original_value
         if original_value == 0
             base_dynamic_gas = {gasPrices|sstoreInitGasEIP2200}
         else
             base_dynamic_gas = {gasPrices|sstoreCleanGasEIP2200}
     else
-        base_dynamic_gas = {gasPrices|sstoreDirtyGasEIP2200}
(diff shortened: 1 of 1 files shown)


2xic (approved):
sorry for the slow review and thanks for your contribution 🙌

PR #342, opened 2024-08-30
Feature/eof opcodes update

2xic:
Built on top of https://github.com/smlxl/evm.codes/pull/339

Only diff is renaming `EOF` -> `EOF (Prague)` and update of solc to a supported version of Prague


dorlevi (approved):
🚢 , I think we need to work on the content a bit more but rather work on it in the open with the help of the community

PR #353, opened 2024-11-09
fix broken url

wangjingcun:
The previous address has been 404, replaced with available

### DIFF
--- a/util/constants.ts
+++ b/util/constants.ts
@@ -1,7 +1,7 @@
 export const GITHUB_REPO_URL = 'https://github.com/smlxl/evm.codes'
 
 // Currently active hardfork from the ones available:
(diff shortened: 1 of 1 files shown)


util/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const GITHUB_REPO_URL = 'https://github.com/smlxl/evm.codes'
 
 // Currently active hardfork from the ones available:
-// See: https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common/src/hardforks
+// See: https://github.com/ethereumjs/ethereumjs-monorepo/tree/stateless/packages/common/src/hardforks
```
2xic, line 4: That isn't the main branch though - I think you mean https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/hardforks.ts ?

wangjingcun:
Friendly ping~

PR #377, opened 2025-03-29
fix: add error case for TSTORE when context is static

### LINKED ISSUE #363: Incomplete description of TSTORE
[TSTORE](https://www.evm.codes/?fork=cancun#5d)
According to the [execution specs](https://ethereum.github.io/execution-specs/src/ethereum/cancun/vm/instructions/storage.py.html) we shouldn't be able to store something in transient storage if we are in a static context.
It should be added to the `Error cases` section in the opcode description.

JereSalo:
Description:
- Specify in the `Error cases` section of the opcode that it will fail if the executed within a static context. 
 
[Execution Specs reference](https://github.com/ethereum/execution-specs/blob/2926c00c1d130e5d0641d278012d42267f3feaf3/src/ethereum/cancun/vm/instructions/storage.py#L171)

Closes #363

### DIFF
--- a/docs/opcodes/5C.mdx
+++ b/docs/opcodes/5C.mdx
@@ -30,3 +30,4 @@ group: Stack Memory Storage and Flow Operations
 The state changes done by the current context are [reverted](#FD) in those cases:
 - Not enough gas.
 - Not enough values on the stack.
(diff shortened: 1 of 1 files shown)


2xic (approved):
Thank you!

PR #414, opened 2026-05-18
Fix API method guards

sueun-dev:
## Summary
- return immediately from API routes when the request method is not supported
- include Allow headers for the supported methods
- avoid calling Etherscan from /api/getContract after rejecting a non-GET request

## Validation
- reproduced the existing /api/getContract behavior: POST returned 200 and still called Etherscan
- verified the patched behavior: unsupported methods return 405 and do not continue handler work
- pnpm run typecheck
- pnpm exec eslint pages/api/getContract.ts pages/api/getDynamicDoc.ts
- pnpm exec prettier --check pages/api/getContract.ts pages/api/getDynamicDoc.ts
- pnpm run build

### DIFF
--- a/pages/api/getContract.ts
+++ b/pages/api/getContract.ts
@@ -7,11 +7,12 @@ export default async function handler(
   res: NextApiResponse,
 ) {
   if (req.method !== 'GET') {
-    res.status(405)
+    res.setHeader('Allow', ['GET'])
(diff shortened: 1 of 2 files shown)


2xic (changes requested):
Why? Have you experienced this problem?

PR #415, opened 2026-05-27
Format playground memory by 32-byte EVM words

NirajBhattarai:
Show memory as offset-labeled rows with continuous hex values and a 0x-prefixed value column so addresses are not confused with byte data.


NirajBhattarai:
fixes #416

PR #423, opened 2026-07-16
fix: clean up spacing in PREVRANDAO documentation

pengqima:
Remove accidental extra spaces from the PREVRANDAO stack output description.

This changes “previous  block's RANDAO  mix” to “previous block's RANDAO mix”, improving readability and keeping the documentation formatting consistent.

### DIFF
--- a/docs/opcodes/44_merge.mdx
+++ b/docs/opcodes/44_merge.mdx
@@ -7,7 +7,7 @@ group: Block Information
 
 ## Stack output
 
-0. `prevRandao`: previous  block's [RANDAO  mix](https://eips.ethereum.org/EIPS/eip-4399).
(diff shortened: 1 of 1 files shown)


pengqima:
@2xic @a-monteiro Hi, Could you please review this PR at your convenience? Thank you very much.

2xic:
@pengqima I don't think this has any user impact as it currently is. What is the benefit of merging this?

pengqima:
> @pengqima I don't think this has any user impact as it currently is. What is the benefit of merging this?

You're right — HTML collapses consecutive whitespace, so this does not change the rendered output. I initially considered the source consistency and readability worthwhile, but I agree that this alone does not provide enough benefit to justify the change.

